PR #79, opened 2026-07-16
Feature/enable tx on dual vfo

fabiojvalente:
# Enable TX on dual VFO + browser TX toggle shortcut

## Summary

This branch adds

* an optional **browser keyboard shortcut for toggling transmit**, 
* fixes several **TX state / split-mode UI issues**
* hardens the **TX-off debounce** in the meter poller.

Context: During this morning while testing my fully remote setup game across a odd behaviour where triggering TX wasn't possible when using split (RX on VFO A and TX on VFO B), did a few small fixes 

Notes: 
* **This was only tested on the FT-DX-10** and am not sure what's the behaviour on dual receiver units
* MR opened against main as this is where I've branched from, please provide guidance if you'd like it against develop instead

### New: configurable TX toggle key (accessibility)
- Added `ApplicationSettings.TxToggleKey` — an optional browser key that toggles transmit. Empty = disabled.
- New Settings UI (`Pages/Settings.cshtml`) with a "Set key" capture flow, a "Clear" button, live capture badge, and a friendly current-shortcut display. Includes a warning that the shortcut directly keys the transmitter.
- Frontend handler in `wwwroot/js/ui/site.js` wires the configured key to the existing `toggleTx()` / `/api/cat/tx` flow. Ignored while typing in an input, textarea, or contenteditable to avoid accidental keying.
- Space handling: a lone `" "` cannot round-trip through an HTML form/input value, so it is persisted as the token `"Space"`. `Settings.cshtml.cs` adds `NormalizeTxToggleKey()` and the required `ModelState.Remove("Settings.TxToggleKey")` (empty is a valid value under `<Nullable>enable</Nullable>`). `Index.cshtml.cs` maps `" "` → `"Space"` and exposes the value via `window.ywcTxToggleKey`.

### Split-mode / inactive-VFO UI fix
- On single-receiver radios, only the **card body** of the inactive VFO is now greyed; the **card header stays full-colour** so the TX button, ZIN, and badges remain clearly active and clickable in split mode (a parent's `opacity`/`filter` could not be undone on nested children before).
- Reworked `.vfo-inactive` CSS to target `.card-body`, with a `.vfo-freq-row` carve-out so the TX VFO's frequency field stays editable in split while other body controls stay read-only.
- `applyVfoActiveStyling()` comments/logic updated accordingly; `toggleTx()` now also calls `updateTxIndicators()`.
- `USER_MANUAL.md` greying-behaviour section updated to match.

### TX-off debounce hardening (`MeterPollingService`)
- Treat both `TX1` and `TX2` as transmitting (mic-keyed on some Yaesu models).
- Accept an authoritative TX-off (web button / voice already cleared `IsTransmitting`) on the first confirming `TX0`, instead of overwriting it back to true for several poll cycles.
- A null poll (radio busy) no longer resurrects a TX state that was already turned off. Debounce threshold factored into `TxOffDebounceCount`.

### CI
- `release.yml` now uploads `Yaesu_Web_Control_Setup.exe` as a workflow artifact on `workflow_dispatch`, and only uploads to a GitHub Release on `release` / tagged dispatch.

## Test plan
- [ ] Set a TX toggle key in Settings (letter, F-key, and Space); confirm it persists across reload and shows the correct current shortcut.
- [ ] Confirm the shortcut toggles TX and is ignored while typing in inputs/text boxes.
- [ ] On a single-receiver radio, verify the inactive VFO body is greyed but the header/TX button stays full-colour and clickable in both normal and split mode.
- [ ] In split mode, confirm the TX VFO frequency field remains editable while other body controls stay read-only.
- [ ] Verify TX-off from the web button clears immediately and is not resurrected by the meter poller.
- [ ] Trigger a manual `workflow_dispatch` and confirm the installer uploads as a workflow artifact.

### home
image

### Settings
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,15 @@ jobs:
           Write-Host "OK: $('{0:N0}' -f (Get-Item Yaesu_Web_Control_Setup.exe).Length) bytes"
         shell: pwsh
 
+      - name: Upload installer as workflow artifact
+        if: github.event_name == 'workflow_dispatch'
+        uses: actions/upload-artifact@v4
+        with:
+          name: Yaesu_Web_Control_Setup
+          path: Yaesu_Web_Control_Setup.exe
+
       - name: Upload installer to GitHub Release
+        if: github.event_name == 'release' || github.event.inputs.tag != ''
         run: |
           # release event carries tag in GITHUB_REF_NAME; dispatch uses the input
           if ("${{ github.event_name }}" -eq "release") {
--- a/Models/ApplicationSettings.cs
+++ b/Models/ApplicationSettings.cs
@@ -167,6 +167,12 @@
         // Yuri W4YSW request 2026-06-17.
         public bool ShowFrequencyArrowButtons { get; set; } = false;
 
+        // Browser key that toggles TX (transmit). Empty string = disabled.
+        // Stored as a KeyboardEvent.key value such as "t", "F8", or " ".
+        // Ignored while typing in inputs to avoid accidental keying during
+        // form entry or frequency editing.
+        public string TxToggleKey { get; set; } = "";
+
         // ── Voice Control (in-process SAPI) ───────────────────────────────
         // When true, the navbar mic button is shown and the SAPI recogniser
         // engages on PTT. Default OFF -- voice control is opt-in so users
--- a/Models/ApplicationSettings.cs
+++ b/Models/ApplicationSettings.cs
@@ -168,7 +168,9 @@
         public bool ShowFrequencyArrowButtons { get; set; } = false;
 
         // Browser key that toggles TX (transmit). Empty string = disabled.
-        // Stored as a KeyboardEvent.key value such as "t", "F8", or " ".
+        // Stored as a KeyboardEvent.key value such as "t" or "F8", except
+        // Space which is stored as the token "Space" (a lone " " cannot
+        // survive HTML form / input value round-trips).
         // Ignored while typing in inputs to avoid accidental keying during
         // form entry or frequency editing.
         public string TxToggleKey { get; set; } = "";
--- a/Pages/Index.cshtml
+++ b/Pages/Index.cshtml
@@ -127,54 +127,47 @@
     #saveMemBtnA, #saveMemBtnB { font-weight: 700 !important; font-size: 0.85rem !important; color: #212529 !important; }
 
     /* Single-receiver radios (FTdx10, FT-710, FTDX3000): the inactive VFO
-       panel and its corresponding spectrum panel are greyed and made
-       non-interactive. Editing any control on the inactive section via
-       CAT would silently swap the radio's active VFO, which is jarring
-       UX — better to disable visually and require an explicit A/B swap
-       on the rig first.
-       Single rule kills ALL pointer interactions (clicks, wheel, hover,
-       focus) so band buttons, frequency-display wheel-tuning, spectrum
-       click-to-tune, and every form control all become inert together.
-       The text/numbers remain VISIBLE — only interaction is suppressed.
+       card BODY is greyed and made non-interactive. The card HEADER stays
+       normal so TX / ZIN / badges remain full-colour and clickable — dimming
+       the whole column (including the header) made the TX button look
+       disabled in split mode even though it was still active (CSS cannot
+       undo a parent's opacity/filter on a child).
+       Editing body controls on the inactive section via CAT would silently
+       swap the radio's active VFO; better to disable visually and require
+       an explicit A/B swap on the rig first.
        See docs/decisions/0003-single-vs-dual-receiver-ui.md
        Reported by Jacek SP3L on #34. */
-    .vfo-inactive {
+    .vfo-inactive > .card > .card-body {
+        pointer-events: none;
+    }
+    .vfo-inactive > .card > .card-body > * {
+        opacity: 0.55;
+        filter: grayscale(0.5);
+        transition: opacity 0.25s ease, filter 0.25s ease;
+    }
+    /* R10/R11 (Jacek SP3L #34): in split mode the inactive panel IS the TX
+       VFO and its frequency must remain interactive so operators can set
+       the TX frequency from YWC without un-splitting first. .vfo-tx-editable
+       is added by applyVfoActiveStyling() only when splitMode>0. The freq
+       row is undimmed as a unit (parent opacity cannot be undone on nested
+       children), then non-frequency siblings in that row (roofing, AF gain)
+       are re-dimmed. */
+    .vfo-inactive.vfo-tx-editable > .card > .card-body > .vfo-freq-row {
+        opacity: 1;
+        filter: none;
+    }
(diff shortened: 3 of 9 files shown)


mm5agm:
Welcome on board, Fabio — really glad to have the first external code contribution on this project, and thanks for taking this on.

I've had a look through the diff (checked against v2.4.2-pre10, the latest) and it looks good overall — the TX2 handling in particular is a nice catch, it actually brings the meter poller in line with something the AI-broadcast handler already assumed. A couple of small things I'd like to sort out before merging:

1. The TX-off debounce threshold changed from 5 to 2 consecutive TX0 readings as part of factoring it into `TxOffDebounceCount`. Was that intentional, or did it just come along for the ride? Just want to make sure we're not reintroducing the "momentary SWR spike" issue the original threshold was guarding against.
2. This PR targets `main` — could you retarget it to `develop`? That's where we land contributions before they go out in a release.

Separately — I recall something in an email that made me think you'd got remote control of YWC working using something called "Mumbles"? Did I read that wrong, or did you actually manage to get remote working going? Would love to hear more if so.

I'm heading out for a while now, so I'll check back for your reply when I'm back. Thanks again!

fabiojvalente:
Hi Colin

Regarding point 1

The Debounce change was trying to fix a behaviour I noticed on my end (maybe induced by possible lan latency), where whenever I disabled the the TX is would take a few seconds to change the button state, while I could hear in the audio that the radio did change the state mostly as soon as I initially clicked the button

Transition was as follows (can try to record a screen cap)

image

This initially led to believe the radio was still TXins when it wasnt

As for Point 2 i've  changed the target to develop and will rebuild locally to confirm things are still fine 


> Separately — I recall something in an email that made me think you'd got remote control of YWC working using something called "Mumbles"? Did I read that wrong, or did you actually manage to get remote working going? Would love to hear more if so.

Transcribing here as I imagine that would be easily lost over the given email :)

> Laptop connected to radio, runs YWC, Mumble Server and Client 
> Then I have a wireguard server that I can use to connect to the home Lan from the outside
> From there on my personal laptop I’m able to connect to the lan address of the box running YWC (and plugged to the radio) and get the audio via mumble

Mumble is a voice chat opensource solution, so you can run your own server and clients

> Mumble config is what took a couple extra minutes to figure out as I had to disable a lot of audio processing to get the most of it, but the main thing is that both ends (personal laptop and audio laptop) are transmitting continuously on mumble and via YWC I’m able to control the radio TX which is at that point just sends/forwards the mumble audio to the antena :) 
> Last part missing is the waterfall, as I don’t have a reliable SDR TX switch, but a poor man’s waterfall can be achieved with the laptop webcam pointing at the yaesu screen :) 

This diagram looks more complex that it really is

```mermaid
flowchart TD
    subgraph Radio["Yaesu Radio"]
        Y1[USB Port]
    end
    subgraph YaesuLaptop["Yaesu Laptop"]
        A2[Yaesu Web Control]
        A3[Mumble Server]
        A4[Mumble Client]
        A3 -- "Receives audio from remote laptop" -->  A4
        A4 -- "Sends audio from radio" --> A3
    end
    A4 -- "Send audio from Remote laptop to Radio IN" --> Radio
    Radio -- "Radio Audio OUT is sent to mumble" --> A4
    A2 -- "Cat Control" --> Radio
    %% Audio_Laptop <-- "CAT and AUDIO IN/OU" --> Radio

    subgraph Remote_Access["Remote Access"]
        B1[WireGuard server]        
    end

    subgraph Personal_Laptop["Personal Laptop"]
        P1[Mumble Client]
        P2[Browser on YWC Page for CAT control]
    end

    Personal_Laptop -- "When outside home" --> Remote_Access --> YaesuLaptop
    Personal_Laptop -- "When at home" --> YaesuLaptop
    P1 -- "Send audio from laptop mic to Mumble" --> A3
    A3 -- "Get audio from radio via mumble" --> P1

 
```


The whole setup was heavily inspired by this [article](https://f6czv.fr/en/remote-access-to-an-ftdx101)

mm5agm:
Hi Fabio,

Thanks for this — it's a genuinely nice piece of work. I went through it properly and the real change is tight and well-judged: the TX-toggle key is opt-in and off by default, the split-mode greying rework (grey the card body, keep the header live) is exactly the right fix for the TX button looking disabled, and the `MeterPollingService` TX-off handling is well reasoned. It follows the existing conventions too — the `ModelState.Remove` for the new optional setting, the `Space`-token round-trip handling, the R8/R10/R11 references. Builds clean my end.

A few things worth flagging:

**Line endings.** Your branch converted two files (`USER_MANUAL.md` and `MeterPollingService.cs`) wholesale from LF to CRLF, which made the raw diff look enormous (~5,900 lines) when the actual change is only ~260. That happened because the repo had no `.gitattributes`, so line endings were left to each person's local git config. I've now added a `.gitattributes` (`* text=auto`) to `develop` that normalizes this automatically. If you pull `develop` and set `git config core.autocrlf true` on your side, your future PRs will stay clean.

**Testing.** You noted you'd only tested on the FT-DX-10, so I ran it on my FTdx101 (dual-receiver). The TX-state changes and the TX-toggle key both check out there. One thing I did catch and fix: with the on-screen frequency keypad open, the TX-toggle key still fired and keyed the rig — the keypad is a `<dialog>` focused on its buttons, so the "ignore while typing" guard didn't recognise it. I've extended that guard to treat an open keypad as an entry surface, which also stops the fullscreen `f` shortcut misfiring there.

**TX confirmation.** I did consider requiring a confirmation before a keyboard-driven transmit, but decided to keep your single-press behaviour — it matches the on-screen TX button, and a confirm step would undercut the point of a quick accessibility shortcut. Good call leaving it simple.

To keep history clean I've integrated your commits onto a normalized branch (with you credited as author/co-author) rather than merging the raw PR, so #79 may show as closed-but-integrated rather than merged — same result, your work's in, just tidied for line endings. `develop` is the right target, by the way — thanks for checking.

73, Colin

mm5agm:
Hi Fabio,

Your changes are live in **v2.4.2-pre11**: https://github.com/mm5agm/Yaesu_Web_Control/releases/tag/v2.4.2-pre11 — the TX-toggle shortcut, the split-mode greying fix, and the TX-off debounce hardening, plus one small follow-up fix I made during testing (the shortcut was still firing while the on-screen frequency keypad was open, so I taught the "ignore while typing" guard to treat the keypad as an entry surface too).

Thanks again for the contribution — good stuff.

73, Colin

mm5agm:
Quick heads-up, Fabio — ignore the "conflicts must be resolved" banner, and please **don't** merge or resolve it. Your changes are already in `develop` (and shipped in pre11); I brought them in as separate commits with the line endings normalized, which is exactly why GitHub now sees a clash on the two files that had CRLF/LF differences plus `site.js` where I added the keypad fix. Nothing needs merging — the PR can just be **closed** whenever you're ready. Thanks again.

fabiojvalente:
Shall I close this MR then ? 
I'll make sure future MRs come with proper line ends :)

mm5agm:
Close please. Thanks for the code.

fabiojvalente:
Closing as the commit were already merged on develop

mm5agm:
Please close.


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

PR #80, opened 2026-07-17
FTdx10 roofing filter support

fabiojvalente:
# Roofing filter CAT control: FTdx10 support + FTDX3000 read-code fix

## Summary

This branch adds working CAT control of the roofing filter on the **FTdx10**,
and corrects the roofing-filter read-back handling on the **FTDX3000**. Both
radios use the same `RF` CAT command as the FTdx101 family, but each has a
model-specific quirk that the previous code got wrong.

- **FTdx10:** roofing filter is now selectable from the VFO panels (it was
  previously assumed to be automatic / uncontrollable). A dead, incorrect `RU`
  handler was removed.
- **FTDX3000:** roofing control already existed but mislabelled 600 Hz / 300 Hz
  and desynced the dropdown in AUTO, because the RF *read* code space differs
  from the RF *set* code space. Read-backs are now normalised.

Verified against the FTdx10 and FTDX3000 CAT Operating Manuals and Hamlib's
`rigs/yaesu/ft3000.c` roofing-filter table.

Screenshot 2026-07-17 at 09 15 25

---

### The bug
The code assumed the FtdX10 selects its roofing filter automatically and offered
no control. It also carried a `case "RU"` handler in `CatMessageDispatcher` that
claimed `RU` was the FtdX10 roofing filter (Auto/15k/6k/3k). Per the manual, `RU`
on the FtdX10 is actually **RX Clarifier Plus Offset** — nothing to do with
roofing — so that handler was dead and wrong.

### The truth (from the manual)
The FtdX10 controls its roofing filter with the `RF` command, single-receiver
style (P1 always `0`):

- Set: `RF0 P2;` — P2 set codes `1`=12 kHz, `2`=3 kHz, `4`=500 Hz, `5`=300 Hz (optional)
- Read: `RF0;` -> `RF0 P3;` — P3 read codes `6`=12 kHz, `7`=3 kHz, `9`=500 Hz, `A`=300 Hz

Note the FtdX10 has **no 1.2 kHz** filter, and its narrow standard filter is
**500 Hz**, not 600 Hz like the FTdx101.

### The fix
- `CatController`: removed FtdX10 from the "selected automatically" early-return
  (FT-710 stays there — it genuinely has no roofing CAT command); added
  `SetFtdx10RoofingFilterAsync` which maps read code -> set code, sends
  `RF0{setCode};`, reads back `RF0;`, and mirrors the result to **both** VFO
  slots (single receiver).
- `CatMessageDispatcher`: deleted the bogus `RU` handler; the `RF` read-back
  path now mirrors to both VFOs on single-receiver radios.
- `Index.cshtml`: roofing dropdown now shown for the FtdX10 (12 kHz / 3 kHz /
  500 Hz standard, 300 Hz optional).
- `Settings.cshtml`: FtdX10 section rewritten — CAT-selectable, single optional
  300 Hz checkbox (value `A`), replacing the previous wrong 1.2 kHz / 300 Hz
  checkboxes.
- `filter-scope-panel.js`: added an FtdX10 roofing Hz map (500 Hz, not 600 Hz).
- `CatMultiplexerService`: init only sends `RF1;` on dual-receiver models.

---

### The bug
The FTDX3000 roofing path treated the RF *read* answer (P3) as if it were the
*set* code (P2). Those two spaces differ on the FTDX3000: 600 Hz is set with `4`
but reads back as `7`, 300 Hz is set with `5` but reads back as `8`, and while in
AUTO the radio reports the filter currently in circuit via `4/5/6/9/A`. The
result: after selecting 600 Hz / 300 Hz, or when in AUTO, the read-back code
didn't match any dropdown value, so the control mislabelled the filter and the
dropdown desynced. The filter-scope panel's `ROOFING_HZ_3000` map was also
shifted/wrong (`3=600`, plus a bogus `5=150`).

### The fix (send path deliberately unchanged)
- New shared table `Services/Ftdx3000Roofing.cs` — single source of truth for the
  set-code names (P2), read-code names (P3), and a read->set normaliser, matching
  the CAT manual and Hamlib's `ft3000.c`.
- `CatController`: both VFO handlers now call `SetFtdx3000RoofingFilterAsync`,
  which still **sends the same `RF0{code};` for codes 0-5 as before** (so 15k/6k/
  3k are byte-for-byte unchanged), then normalises the read-back to the dropdown
  set-code space and mirrors to both VFO slots.
- `CatMessageDispatcher`: single-receiver `RF` read-backs are normalised for the
  FTDX3000 (via a new `RadioModel` field on `RadioStateService`) so init/AI
  broadcasts of 600/300/AUTO no longer desync the dropdown. FtdX10 is untouched.
- `RadioStateService` / `RadioInitializationService`: added a `RadioModel` field,
  set at startup, so the dispatcher can distinguish FTDX3000 from other
  single-receiver radios.
- `filter-scope-panel.js`: `ROOFING_HZ_3000` corrected to match the dropdown set
  codes (`1=15k, 2=6k, 3=3k, 4=600, 5=300`); Auto (`0`) draws no roofing outline.

### Not changed
- The FTDX3000 send codes for `1`/`2`/`3` (and all others) are untouched, so the
  common case cannot regress.
- No FTDX3000 dropdown option changes.

---

## Files changed
- `Controllers/CatController.cs`
- `Services/CatMessageDispatcher.cs`
- `Services/RadioStateService.cs`
- `Services/RadioInitializationService.cs`
- `Services/CatMultiplexerService.cs`
- `Services/Ftdx3000Roofing.cs` (new)
- `Pages/Index.cshtml`
- `Pages/Settings.cshtml`
- `wwwroot/js/ui/filter-scope-panel.js`

### DIFF
--- a/Controllers/CatController.cs
+++ b/Controllers/CatController.cs
@@ -717,19 +717,22 @@ namespace Yaesu_Web_Control.Controllers
             { "A", "5" }   // 300 Hz (option)
         };
 
-        // FTdx10 roofing filter display names (RU command code -> display name)
+        // FTdx10 roofing filter display names (RF read code P3 -> display name)
         private static readonly Dictionary<string, string> FtdxTenRoofingFilterNames = new()
         {
-            { "1", "15 kHz" },
-            { "2", "6 kHz" },
-            { "3", "3 kHz" }
+            { "6", "12 kHz" },
+            { "7", "3 kHz" },
+            { "9", "500 Hz" },
+            { "A", "300 Hz" }
         };
 
-        // FTDX3000 roofing filter display names (RF0x code -> display name)
-        private static readonly Dictionary<string, string> Ftdx3000RoofingFilterNames = new()
+        // FTdx10 roofing filter set codes (read code P3 -> set code P2 used in RF command)
+        private static readonly Dictionary<string, string> FtdxTenRoofingFilterSetCodes = new()
         {
-            { "0", "Auto" }, { "1", "15 kHz" }, { "2", "6 kHz" },
-            { "3", "3 kHz" }, { "4", "600 Hz" }, { "5", "300 Hz" }
+            { "6", "1" },  // 12 kHz
+            { "7", "2" },  // 3 kHz
+            { "9", "4" },  // 500 Hz
+            { "A", "5" }   // 300 Hz (optional)
         };
 
         [HttpPost("roofingfilter/a")]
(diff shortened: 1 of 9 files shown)


fabiojvalente:
Good morning@mm5agm 
Quick note: The ftdx3000 changes where based on the CAT control manual and I'm not able to confirm them, the FTDX10 ones however are working as expected :)

mm5agm:
Hi Fabio,

Just merged #80 — thank you. The roofing-filter work is excellent: properly researched against the CAT manuals and Hamlib, the FTDX3000 read-vs-set code fix is exactly right, and getting the FTdx10 roofing filter actually controllable is a real improvement. That's now on `develop`.

More generally — your contributions have been consistently thoughtful and well-tested (#79 and now #80), and you keep spotting genuine issues. It's genuinely appreciated, and I'd like to make it easier for you to work on YWC: I've **added you as a collaborator** on the repo. You should see an invitation. Once you accept, you can push branches straight to the repo and open PRs from them without needing your own fork. I've kept branch protection on `develop`/`main`, so everything still comes in via PR and I'll review/merge as normal — nothing changes there, it just saves you the fork step.

One thing I'd like to do: **credit you properly in the README.** Could you let me know your **callsign** (and how you'd like your name to appear) so I can add you to a contributors section?

Thanks again — really glad to have you involved.

73, Colin MM5AGM

Issue #35, opened 2026-06-14
[BUG] RF Power setting is not synchronized to the radio after starting YWC


SP3L-Jacek:
## Steps to reproduce

1. With YWC connected to the radio, set RF Power to some arbitrary value, for example to 30 W.
2. Close YWC (that is close the tab in the internet browser and select "Exit Yaesu Web Control" after right-clicking on the YWC icon sitting in the Windows tray).
3. Manually change RF setting on the radio front panel, for example to 10 W.
4. Open YWC.
5. Observe that it shows the RF power value that was valid before its closure (30 W in the above example).

## Expected behaviour

YWC should synchronize to the RF Power set in the radio.

## Actual behaviour

YWC remembers the RF Power setting it had right before application closure.

## Diagnostics

```
YWC Version    2.3.6 (released 2026-06-12)
Radio model    FTdx10
Band plan      Region1
Serial port    COM7 at 38400 baud
Connection     connected
SDR device     (none configured)
DX cluster     (off)
Cluster login  (blank)
Browser        Microsoft Edge 149
.NET runtime   10.0.8
Operating sys  Microsoft Windows 10.0.26200
```

## Anything else?

Any change in RF power setting, made either in the radio, or in the application, synchronizes them.

mm5agm:
Jacek — both #35 (RF Power not synced on startup) and #36 (slider doesn't follow front-panel change) are fixed on develop and will ship in v2.3.7. While testing the #35 fix I noticed a related rounding edge case (radio at 91 W → YWC label was showing 90 W because of the slider's step=5 snap), so I've made the label show the exact radio value while the slider visually snaps to the nearest step — should keep operators using odd watt values like 91 / 73 happy.

Plus the "100W" → "100 W" space-before-unit fix from your closing note in #36.

73, Colin MM5AGM

SP3L-Jacek:
Bug still present in version 2.3.7.

mm5agm:
Jacek — the fix is in v2.3.7 (commit 89326ca; `RadioInitializationService` no longer writes the persisted Power on connect, and queries `PC;` from the radio on every connect instead). Could you describe what you're seeing now in the same Steps to reproduce format as the original report? Specifically:

1. Power value you set on the radio's front panel **before** starting YWC
2. Power value YWC's slider shows **immediately after** startup
3. Power value the radio actually has **after** YWC starts (read on the radio itself — not via YWC)

[...]

Issue #16, opened 2026-06-04
[BUG] YWC overwrites the MIC GAIN and PROC LEVEL settings in the radio after the first connection


SP3L-Jacek:
## Describe the bug

MIC GAIN changed from 33 to 50 and PROC level changed rom 100 to 50.

## Steps to reproduce

1. Install the YWC and connect to the radio (FT10DX10 in this case).
2. Check the MIC GAIN and PROC LEVEL settings.
3.

## Expected behaviour

I expect them to stay as originally set before connecting to YWC.

## Actual behaviour

They changed

## Diagnostics

```
YWC Version    2.1.0 (released 2026-06-03)
Radio model    FTdx10
Band plan      Region1
Serial port    COM7 at 38400 baud
Connection     connected
SDR device     (none configured)
DX cluster     cluster.dl4ny.de:7300
Cluster login  SP3L
Browser        Microsoft Edge 148
.NET runtime   10.0.8
Operating sys  Microsoft Windows 10.0.26200
```

mm5agm:
Hello Jacek, this is a real bug and a clear one — thank you.

**Root cause:** `Services/RadioInitializationService.cs` was *writing* persisted state back to the radio on every connect for **MIC GAIN**, **Speech Processor on/off**, and **PROC LEVEL**. On a fresh YWC install with no prior persisted values, those default to roughly `50` / off / `50`, so on first connect YWC was happily sending `MG050;`, `PR0;`, and `PL050;` to your radio — overwriting your **MIC GAIN = 33** and **PROC LEVEL = 100**. The 50 / 50 / 50 you saw afterwards was YWC's defaults applied as if you had set them.

**Why this design was wrong:** MIC GAIN and PROC LEVEL are about TX audio quality. They're normally tuned physically on the radio for the operator's voice and mic, not via YWC. A remote app shouldn't be silently overwriting them on every startup.

**Fix on develop:** YWC no longer writes `MG`, `PR`, or `PL` on connect. The radio is now the source of truth for those three values. The existing `MG;`, `PR;`, and `PL;` query commands in the connect sequence already populate YWC's UI sliders with whatever the radio actually has. If you adjust a value via a YWC slider afterwards, that sends the corresponding command to the radio (and the radio's state changes accordingly) — but YWC will no longer silently apply a saved value on connect.

[...]

PR #81, opened 2026-07-17
Single-receiver VFO routing fixes: memory recall, roofing filter, A/B re-query + full state replay for late-joining clients

fabiojvalente:
# Single-receiver VFO routing fixes: memory recall, roofing filter, A/B re-query + full state replay for late-joining clients

TL,DR: Fix some state drift between what the radio have and what YWC reports

Add labels in the VFOs for `band` and `mode`
image

## Summary

On single-receiver radios (FTdx10, FT-710, FTDX3000, FT-991A), VFO A and B are frequency/mode slots for one receiver; the operating VFO is tracked by `ActiveVfo` (from the `VS` CAT command). Several code paths still ignored that and hard-wired VFO A. This branch fixes them, plus a related state-sync bug for browsers that connect after startup.

### Memory recall tunes the active VFO (`Controllers/MemoryController.cs`)

- Recall (`POST /api/memory/{id}/recall`) was hard-coded to VFO A — always sending `FA`/`MD0` and writing only `*A` state fields. Recalling a memory while VFO B was active updated the wrong slot, so the receiver did not move.
- `Recall` now uses `RadioCapabilities.VfoIsB` — the same routing helper used throughout `CatController` — to target the active VFO. Mode and frequency commands switch between `MD0`/`FA` and `MD1`/`FB`, and every per-VFO state write (clarifier, antenna, IF width/shift, roofing, NB, NR, AGC) routes to the matching A or B property. Dual-receiver behaviour is unchanged (recall still targets VFO A).

### Roofing filter is per-VFO state on single-receiver radios (`CatController`, `CatMessageDispatcher`, `RadioInitializationService`)

- The FTdx10/FTDX3000 roofing setters and the `RF` read-back handler used to mirror the filter code to both `RoofingFilterA` and `RoofingFilterB`. But the radio stores the roofing filter per VFO slot, so mirroring corrupted the inactive slot's value.
- Setters now write only the active VFO's slot, and the dispatcher's `RF` case routes through `SetPerVfo` like the other P1=0-Fixed receive controls (including the FTDX3000 read-code normalisation). `RF0;` was added to the init per-VFO query list so both slots are populated at connect.

### `SetMode` consolidation (`CatController`)

- Refactored to build the CAT command from `VfoP1Outgoing` and route state writes via `VfoIsB` instead of duplicated A/B branches — same fix pattern as the other receive-control endpoints, and the Contour/APF re-apply after a mode change now reads the correct per-VFO state.

### Front-panel A/B press re-syncs per-VFO controls (`CatMessageDispatcher`, `CatMultiplexerService`, `CatCommands`)

- Pressing A/B on the rig fires a burst of P1=0-Fixed auto-info broadcasts racing the `VS` answer. The 300 ms dispatch buffer previously applied any queued update to whichever `ActiveVfo` was current at flush time — updates spanning the VS boundary could land in the wrong slot. Buffered updates whose `ActiveVfo` changed between enqueue and flush are now dropped as ambiguous.
- Ground truth is restored by a new debounced post-VS re-query: 500 ms after a `VS` change (cancel-and-restart on rapid flips), the multiplexer re-reads the full per-VFO control set for the newly active VFO. The query list is now shared as `CatCommands.SingleReceiverPerVfoQueries` between init ping-pong and the re-query. `MD` read-back also routes via `SetPerVfo` now.
- AF Gain persisted-state restore at init is skipped on single-receiver radios — the radio is the source of truth there (same precedent as Mode #38, RF Power #35, MIC Gain #16); the `AG0;` queries populate the UI instead.

### Late-joining browsers get the full radio state (`RadioHub`, `RadioStateService`)

- SignalR broadcasts only fire on *change*, and `OnConnectedAsync` replayed only `FrequencyA`/`FrequencyB`. A browser connecting after startup (second tab, another computer) kept the frontend JS defaults for anything not server-rendered in the Razor page — most visibly `ActiveVfo`/`TxVfo`/`SplitMode`, which made VFO A always look active on late-joining clients.
- `RadioStateService.GetClientStateSnapshot()` returns every UI-relevant property (meters excluded), and the hub replays it to the connecting client through the normal `RadioStateUpdate` envelope. No frontend changes needed; voice announcements are unaffected (initial-load suppression already swallows first values).

## Test plan

- [x] On FTdx10: with VFO B active, recall a memory — VFO B retunes and the UI updates
- [x] On FTdx10: with VFO A active, recall a memory — VFO A retunes (regression)
- [ ] On FTdx10: press A/B on the rig — after ~0.5 s all per-VFO controls (mode, roofing, AGC, IF width/shift, NB/NR, contour/APF, AF/RF gain, squelch) show the newly active VFO's values
- [ ] On FTdx10: change the roofing filter with VFO B active, press A/B — each panel keeps its own filter (no mirroring)
- [ ] On FTdx10: with VFO B active, open the UI in a second browser/computer — VFO B shows as active; split/TX state also correct
- [ ] On FTdx101 (dual receiver): memory recall still targets VFO A; mode/roofing per-VFO commands unchanged; second-browser state still correct

### DIFF
--- a/Controllers/CatController.cs
+++ b/Controllers/CatController.cs
@@ -875,7 +875,8 @@ namespace Yaesu_Web_Control.Controllers
 
         /// <summary>
         /// FTdx10 single-receiver roofing filter: RF0 P2 set / RF0 P3 read.
-        /// Mirrors the result to both VFO slots so both panels stay in sync.
+        /// Per-VFO state is tracked in the active VFO slot (inactive panel is
+        /// not editable on single-receiver radios).
         /// </summary>
         private async Task<IActionResult> SetFtdx10RoofingFilterAsync(RoofingFilterRequest request)
         {
@@ -893,8 +894,8 @@ namespace Yaesu_Web_Control.Controllers
             if (!string.IsNullOrEmpty(rfReadResponse) && rfReadResponse.Length >= 4)
             {
                 var actualFilter = rfReadResponse[3].ToString();
-                _radioStateService.RoofingFilterA = actualFilter;
-                _radioStateService.RoofingFilterB = actualFilter;
+                if (_radioStateService.ActiveVfo == 1) _radioStateService.RoofingFilterB = actualFilter;
+                else                                   _radioStateService.RoofingFilterA = actualFilter;
 
                 if (actualFilter != request.Filter)
                 {
@@ -909,8 +910,8 @@ namespace Yaesu_Web_Control.Controllers
                 return Ok(new { message = $"Roofing filter {filterName} selected", filter = actualFilter, filterName });
             }
 
-            _radioStateService.RoofingFilterA = request.Filter;
-            _radioStateService.RoofingFilterB = request.Filter;
+            if (_radioStateService.ActiveVfo == 1) _radioStateService.RoofingFilterB = request.Filter;
+            else                                   _radioStateService.RoofingFilterA = request.Filter;
             var fallbackName = FtdxTenRoofingFilterNames.GetValueOrDefault(request.Filter, request.Filter);
             return Ok(new { message = $"Roofing filter {fallbackName} selected", filter = request.Filter, filterName = fallbackName });
         }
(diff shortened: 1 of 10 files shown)


fabiojvalente:
@mm5agm for credit purposes my call is CR7CDC

fabiojvalente:
@mm5agm found a possible regression, fixing it right now

mm5agm:
Thanks Fabio! Just so I get the credit exactly right in the README — could you confirm the callsign spelling? I couldn't turn up CR7CDC on QRZ, though I know plenty of people don't keep a profile there, so it may simply not be listed. Just want to have it letter-perfect before it goes in.

73, Colin MM5AGM

fabiojvalente:
@mm5agm this one got rather bigger than anticipated
During testing I've discovered a few scenarios where there was state drift between YWC and the radio, and some misbehaviour when changing settings on VFO B that were carried over to VFO A when switching

I dont think dual VFO units are affected, but testing is recommended

As I'm using YWC for remote operation over Lan, I also noticed YWC would show VFO A as active (when B was in fact the active one) and would never get the correct info form the radio
This MR also fixes this as it considers the radio to be the souce of truth and polls this info on a recurring basis

As for QRZ, I've created the acc and waiting now to be added 👍

PR #92, opened 2026-08-05
Remote audio streaming (browser ↔ radio USB) + optional HTTPS

fabiojvalente:
⚠️ This is experimental, confirmed to work with a FTDX10 on MacOS :)

# Native remote audio (browser ↔ radio USB) + optional HTTPS

Tl,DR: This PR provides audio streaming to YWC, removing the requirement to fiddle with mumble or sonobus.
 It was designed to be as frictionless as possible, 
To test it in its current state: 
1. https needs to be enabled (is needed for the mic), the user can generate a self signed certificate in the settings
2. user must select the audio in and out interfaces for the radio in the settings
3.  Allow the browser to access the mic
4. Click Start Audio

## Summary

Adds **in-browser send/receive audio** so a remote operator on LAN/VPN can operate SSB without Mumble/SonoBus. Built on top of `feature/macos-and-linux-build` (this MR is **audio/HTTPS only** — no platform/Docker changes).

- **Remote Audio bridge:** PortAudio host I/O ↔ dedicated `/audio` WebSocket ↔ browser mic/speakers (PCM16 preferred; Opus via Concentus / WebCodecs when negotiated).
- **Settings:** opt-in Remote Audio section with radio RX (capture) / TX (playback) device pickers and gains; devices open only while a session is active.
- **Single session:** second browser is rejected busy; PTT stays on existing CAT TX controls.
- **Optional self-signed HTTPS:** generate cert in Settings, dual-listen HTTP + HTTPS after restart — required for remote `getUserMedia` (secure context).
- **Docs:** USER_MANUAL §6.2 / §6.8 / §18 and FAQ §15.7 updated.
image
image

### Notable code

| Area | Change |
|------|--------|
| `Services/Audio/*` | Bridge, PortAudio enumeration, Opus codec, wire protocol, HTTPS cert helper |
| `Controllers/AudioController.cs` | `/api/audio/devices`, status, cert generate |
| `Program.cs` | WebSockets `/audio`; Kestrel dual-bind when HTTPS enabled |
| `Pages/Settings.cshtml` | Remote Audio + HTTPS UI |
| `Pages/Index.cshtml` + `wwwroot/js/audio/*` | Start/Stop bar, mute, levels |
| `Models/ApplicationSettings.cs` | `AudioStreaming*`, `Https*` settings |
| `USER_MANUAL.md` | Remote audio setup, MOD SOURCE, TLS/VPN troubleshooting |

## Test plan

- [x] `dotnet build -f net10.0` and `-f net10.0-windows` succeed
- [x] Settings → Remote Audio: enable; RX/TX device lists populate (Refresh works); Save persists
- [x] Index Remote Audio bar appears only when enabled; Start prompts for mic; RX audible; TX meter moves when speaking
- [x] CAT TX button / toggle still keys the radio independently of the audio stream
- [x] Second browser Start audio → busy / rejected
- [x] Disable Remote Audio → bar hidden; no PortAudio devices held
- [x] HTTPS: Generate cert with SAN (LAN/WG IP); Enable HTTPS; Save; **restart**; open `https://…:8443`; accept warning; Start audio works remotely
- [x] Radio MOD SOURCE = USB/REAR; confirm TX audio modulates when keyed
- [ ] Voice Control mic (Windows) remains independent of radio USB device pickers
- [x] Diff vs `feature/macos-and-linux-build` contains only audio/HTTPS files (no Docker/macOS tray churn)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,16 +175,28 @@ jobs:
         run: |
           image="ghcr.io/$(echo '${{ github.repository }}' | tr '[:upper:]' '[:lower:]')"
           echo "image=${image}" >> "$GITHUB_OUTPUT"
+          # :latest only moves on full releases — never on pre-releases or smoke builds.
+          latest=false
           if [ "${{ github.event_name }}" = "release" ]; then
             echo "version=${GITHUB_REF_NAME}" >> "$GITHUB_OUTPUT"
             echo "push=true" >> "$GITHUB_OUTPUT"
+            if [ "${{ github.event.release.prerelease }}" != "true" ]; then
+              latest=true
+            fi
           elif [ -n "${{ github.event.inputs.tag }}" ]; then
-            echo "version=${{ github.event.inputs.tag }}" >> "$GITHUB_OUTPUT"
+            version="${{ github.event.inputs.tag }}"
+            echo "version=${version}" >> "$GITHUB_OUTPUT"
             echo "push=true" >> "$GITHUB_OUTPUT"
+            # SemVer pre-release (e.g. v2.4.3-pre6) has a hyphen after the numeric version.
+            ver="${version#v}"
+            if [[ "$ver" != *-* ]]; then
+              latest=true
+            fi
           else
             echo "version=smoke" >> "$GITHUB_OUTPUT"
             echo "push=false" >> "$GITHUB_OUTPUT"
           fi
+          echo "latest=${latest}" >> "$GITHUB_OUTPUT"
 
       - name: Log in to GHCR
         if: steps.prep.outputs.push == 'true'
@@ -202,7 +214,7 @@ jobs:
           tags: |
             type=semver,pattern={{version}},value=${{ steps.prep.outputs.version }},enable=${{ steps.prep.outputs.push }}
             type=semver,pattern={{raw}},value=${{ steps.prep.outputs.version }},enable=${{ steps.prep.outputs.push }}
-            type=raw,value=latest,enable=${{ steps.prep.outputs.push }}
+            type=raw,value=latest,enable=${{ steps.prep.outputs.latest }}
             type=raw,value=smoke,enable=${{ steps.prep.outputs.push != 'true' }}
 
       - name: Build and push
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ appsettings.user.json
 .cursor
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
-
+.DS_Store
 # Build results
 [Dd]ebug/
 [Dd]ebugPublic/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,7 +46,7 @@ On macOS, set **Serial Port** to a `/dev/cu.*` device. On Linux, use `/dev/ttyUS
 | Serial validation | `COM*` | `/dev/…` |
 | AppData | `%APPDATA%\MM5AGM\Yaesu Web Control\` | `$XDG_CONFIG_HOME` or `~/.config/MM5AGM/Yaesu Web Control/` |
 | `AutoShutdownWhenNoBrowsers` | Default true | Default true; `HostRuntime.IsContainer` forces keep-alive |
-| Browser auto-open | Yes | Skipped when `HostRuntime.IsContainer` |
+| Browser auto-open | `OpenBrowserOnStartup` (default true) | Same setting; skipped when `HostRuntime.IsContainer` |
 | Operator-facing docs | `USER_MANUAL.md` §§1–4, 6.2–6.3, 15.10, 17 | Same |
 
 There are no automated tests. Verification is manual via the browser at `http://localhost:8080`.
--- a/Controllers/AudioController.cs
+++ b/Controllers/AudioController.cs
@@ -31,9 +31,27 @@ namespace Yaesu_Web_Control.Controllers
             try
             {
                 var inputs = AudioDeviceEnumerator.ListInputs()
-                    .Select(d => new { d.Index, d.Name, d.DefaultSampleRate });
+                    .Select(d => new
+                    {
+                        d.Index,
+                        d.Name,
+                        d.HostApiName,
+                        d.HostApiIndex,
+                        displayName = d.DisplayName,
+                        key = d.PersistenceKey,
+                        d.DefaultSampleRate
+                    });
                 var outputs = AudioDeviceEnumerator.ListOutputs()
-                    .Select(d => new { d.Index, d.Name, d.DefaultSampleRate });
+                    .Select(d => new
+                    {
+                        d.Index,
+                        d.Name,
+                        d.HostApiName,
+                        d.HostApiIndex,
+                        displayName = d.DisplayName,
+                        key = d.PersistenceKey,
+                        d.DefaultSampleRate
+                    });
                 return Ok(new { inputs, outputs });
             }
             catch (Exception ex)
--- a/Controllers/AudioController.cs
+++ b/Controllers/AudioController.cs
@@ -1,0 +1,94 @@
+using Microsoft.AspNetCore.Mvc;
+using Yaesu_Web_Control.Services;
+using Yaesu_Web_Control.Services.Audio;
+
+namespace Yaesu_Web_Control.Controllers
+{
+    [ApiController]
+    [Route("api/audio")]
+    public class AudioController : ControllerBase
+    {
+        private readonly ISettingsService _settings;
+        private readonly AudioSessionManager _sessions;
+        private readonly RadioAudioBridgeService _bridge;
+        private readonly ILogger<AudioController> _logger;
+
+        public AudioController(
+            ISettingsService settings,
+            AudioSessionManager sessions,
+            RadioAudioBridgeService bridge,
+            ILogger<AudioController> logger)
+        {
+            _settings = settings;
+            _sessions = sessions;
+            _bridge = bridge;
+            _logger = logger;
+        }
+
+        [HttpGet("devices")]
+        public IActionResult ListDevices()
+        {
+            try
+            {
+                var inputs = AudioDeviceEnumerator.ListInputs()
+                    .Select(d => new { d.Index, d.Name, d.DefaultSampleRate });
+                var outputs = AudioDeviceEnumerator.ListOutputs()
+                    .Select(d => new { d.Index, d.Name, d.DefaultSampleRate });
+                return Ok(new { inputs, outputs });
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Failed to enumerate audio devices");
+                return StatusCode(500, new { error = ex.Message });
+            }
+        }
+
+        [HttpGet("status")]
+        public async Task<IActionResult> Status()
+        {
+            var s = await _settings.GetSettingsAsync();
+            return Ok(new
+            {
+                enabled = s.AudioStreamingEnabled,
+                sessionActive = _sessions.HasActiveSession,
+                devicesOpen = _bridge.DevicesOpen,
+                codec = _bridge.ActiveCodec,
+                rxLevel = _bridge.RxLevel,
+                txLevel = _bridge.TxLevel,
+                httpsEnabled = s.HttpsEnabled,
+                httpsPort = s.HttpsPort,
+                certificatePresent = HttpsCertificateService.CertificateExists,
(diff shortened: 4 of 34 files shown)


fabiojvalente:
Have updated the ui experience

Use icons with tooltips

Screenshot 2026-08-06 at 08 42 30

User can popout audio so the stream doesnt stop if the user navigates to a different page (ex settings), this can also be extended in the future to control audio settings in the radio, ex mic gain proc etc
image


And given we now get the audio in the browser, the vfo scope will now show the true audio when the audio is connect, otherwise fallback to Math.random 
@mm5agm this needs to be check and likely adjusted for radios with dual VFO

With audio disconnected
image

With audio connected
image

mm5agm:
Hi Fabio,

First off - this is brilliant. I've had it running on my FTdx101MP: I can hear the radio in the browser on RX, and the TX capture meter moves when I speak, so the core pipeline is clearly working end to end. Really pleased with it.

A couple of things came up while testing that might be worth a thought - no rush, I know it's still a draft.

**1. Duplicate device names make picking the right device hard**

On Windows the same USB CODEC shows up many times over - once per audio subsystem (MME / WASAPI / DirectSound / WDM-KS) - so the RX and TX dropdowns are full of identically-named entries and it's trial-and-error to find the one that actually carries the radio audio. Because a device is saved and matched by its name string, there's no way to tell the duplicates apart. Could we show something alongside the name to disambiguate - the PortAudio host API and/or the device index - and persist that too? Even just appending the host API in the dropdown would help a lot.

**2. Background buzz/hiss on RX - maybe an optional noise gate?**

There's a steady background buzz and hiss under the received audio. I've chased it down and I'm fairly sure it isn't your code - it's a fixed noise floor in the radio's USB-CODEC output stage:

- it doesn't change with RF Gain,
- it's still there with no station at all (RF Gain at 0),
- and it's independent of the radio's per-mode OUT LEVEL menu (that only affects the rear RTTY/DATA jack, not the USB audio).

So YWC is faithfully passing along a hiss that's already in the USB stream before it reaches the browser. The *buzz* part specifically looks like USB bus noise / a ground loop, which I can chase at my end with an isolator or ferrite.

Where YWC might help: an optional RX noise gate / squelch that mutes the output during quiet passages (SSB/CW gaps, idle) so the buzz isn't just sitting there. It won't help a continuous AM broadcast where there's always a carrier, but for SSB/CW it would clean things up nicely. Feasible in the audio path?

That's it for now - genuinely impressed. Happy to keep testing whatever's useful.

73 Colin

fabiojvalente:
Thank you so much for the input @mm5agm 

> 1. Duplicate device names make picking the right device hard
I agree, however I only have a FTDX10 at hand and cannot replicate that scenario on my own easily. 
Nevertheless I will try to get some unique id per device to distinguish them in the ui. 
Same problem might occur on macos and linux setups

About the noise I didn't notice it here (but will try to replicate with different hardware)
Sadly the propagation has been dead on 11/10m (am not allowed to operate outside those yet :( ) for a few days and I haven't been able to check how my audio is reaching nearby SDRs 

Neverthesless the rx noise filtering might be a good addon on poorly grounded machines

fabiojvalente:
Windows duplicate devices  is now fixed: Same USB CODEC showed up under MME / DirectSound / WASAPI / WDM-KS with identical names. Dropdowns now label devices with the host API and persist name [hostApi].

Pop-out stop left Home greyed out — Stopping in the pop-out kept “ownership” on the main bar so Start/Stop/mutes stayed disabled. Ownership now only applies while the pop-out is actually streaming.

Browser mic picker — Mic dropdown on Home and pop-out so you can choose which browser microphone goes to the radio; choice is remembered in the browser.

mm5agm:
If you are finished with this it can be merged Fabio - users of IWC, the icom clone of YWC have been asking if it was possible to implement this . Could it be ported ? THey want to be able to run IWC in another room to where the radio is.

fabiojvalente:
Hi Colin,

To start with, I don't have a icom unit with, so am completely unfamiliar with how it would work, however I don't imagine it would be too much trouble, assuming it exposes audio interface like the yaesu does, then its only that bit thats missing (we can already do ptt/mox over CAT)

As being finished with this, this specific branch is blocked by the other PR
**Missing on this one**

- Add/bundle RX/TX gain cotnrols in the main operation view instead of the settings


**On the upstream branch (macos-and-linux-build)**
- Test the actions to build and ship the distributables for MacOS and Linux


Ideally, I hope to have both ready to ship into develop by End of week

mm5agm:
Thanks Fabio - that all matches what I found digging through it. The IC-7300 exposes a USB audio CODEC just like the FTdx10 (one cable carries both CI-V and the sound card), and IWC already keys the radio over CI-V, so PTT's covered. The audio bridge itself doesn't touch CAT at all - `Services/Audio/*` only depends on PortAudio and the settings service, nothing radio-specific - so you're right that it's a small gap to close.

That's actually what makes me want to raise the bigger point before either of us starts copying files. Because the audio subsystem is completely radio-agnostic, if I just paste it into IWC we'll have two copies to maintain forever - every fix you make here (the duplicate-device-name work, a future noise gate, and so on) would have to be redone by hand on the Icom side. That's exactly the trap.

I'm already planning to pull the radio-agnostic voice code into a **shared library** that both YWC and IWC consume from one source. The audio bridge looks like an ideal second tenant for that same library - the only per-app parts are the Settings/Index UI and the DI wiring, which are thin. So both apps would build against one audio codebase and your fixes land in both at once.

No urgency, and I don't think it changes anything about this PR - I'd let this merge and settle in YWC first, then lift the finished version, with your permission, into the shared lib rather than port a moving target. But I wanted to flag the direction now in case it's worth keeping the seam clean (e.g. no accidental coupling to YWC-specific types) while it's fresh. Does that fit how you'd want to structure it? Keen to hear your thoughts.

73 Colin

fabiojvalente:
That does make sense 👍

fabiojvalente:
@mm5agm This is now ready for review and merging if you're happy,

## Today's progress

- Reorganised ui for group mic and gain settings
- User can now control from the operation/home view the browser audio and mic gain
- Default audio codec to Opus, as the bitrate is lower, for SSB the difference is unnoticeable from my tests. In any case the user can still switch to PCM for uncompressed audio. The differences are `32kbs vs 768kbs` each way, or `64 kbps vs 1536kbps` for RX and TX
- Am also piggybacking in this MR the changes to the release workflow for container images, so latest only gets updated on `stable` releases

### Screenshots

image
image

mm5agm:
Hi Fabio,

Bench-tested this on my FTdx101MP (dual VFO) on Windows this evening. Good news first: RX audio works, TX works, and the "ringing/feedback" I was worried about turned out to be my own config - I had left the TX device blank, so the bridge sent my browser mic to my monitor speakers. Setting TX to the radio's USB Audio CODEC speakers fixed it completely. Might be worth a small guard so a blank/non-radio TX device can't be selected by default, but that is a nice-to-have, not a blocker.
image

The one item you flagged for dual-VFO does reproduce. With audio streaming, the passband scope shows the same live audio trace on BOTH VFO A and VFO B (screenshot below). Since the USB audio is mono from the Main receiver (VFO A on the 101MP), VFO B's scope is really showing VFO A's audio. It should drive only the VFO the audio belongs to, with the other VFO's scope on its normal fallback. You know the audio routing better than I do (and whether B's audio can be selected at all), so I'll leave the exact approach to you.

Everything else passed. Once the scope is tied to the correct VFO I'm happy to merge.

(Unrelated aside, not from this PR: I also have a separate dual-SDR spectrum issue where VFO A's panadapter freezes while B updates. I've confirmed that is pre-existing and nothing to do with your audio changes, so please ignore it here - I'll chase it separately.)

Thanks again for all of this.
Colin

fabiojvalente:
That’s great to hear !
  

  

  
As I don’t have a dual receiver unit, can you confirm if it exposes one single audio interface that contains the audio from both or the computer is able to see two interfaces ?
  
If the first option, then there’s isn’t much we can do about it
  
If the second, then we need to add a option to define audio settings for the second receiver   
  

  
Can you send a screenshot (to my personal email if you need) of the remote audio section in the settings page with the drop down listing what interfaces you see ?
  
  
  
  
  
  

  
  
>

mm5agm:
[cid:59ccabea-0692-40c5-a43f-142e66135ac2]
Screen dump[cid:d6997ba3-ae1e-4b86-8db0-ce4ed3762cf4]
Screen dump for RX and TX on settings page.
Colin MM5AGM
________________________________

fabiojvalente:
Colin, I've been doing some research on this
First, thanks for the thorough bench test on the 101MP — really helpful, and good to hear RX/TX are solid and that the ringing was the blank TX device (monitor speakers). **I’ll add a guard so a blank/non-radio TX device can’t sit as the default; that’s a nice follow-up, not in this PR.**

**On the dual-VFO filter scopes**: you’re right that both panels show the same live trace today. The radio exposes a single USB Audio CODEC (not one device per VFO). On the 101MP that path is typically stereo (MAIN/SUB on L/R), but the current implementation downmixes to mono for the Remote Audio bridge and then attaches that one FFT to both VFO A and VFO B scopes — so B’s live bars are misleading (and the browser sees only the MONO stream).

I’ll treat correct per-VFO scoping (and audio) as a separate MR given the work (and blind testing that I cannot do given the lack of hardware) involved after this merges, my suggestion is:
- first stop putting the live FFT on VFO B (decorative fallback there while audio is mono), 
- then keep stereo through the pipe on dual-receiver radios and map L→A / R→B for true dual traces (with a bench check of the channel map).

Let me know your thoughts

Thanks,
Fabio

mm5agm:
Thanks Fabio, that's a really clear diagnosis and it matches exactly what I saw on the bench - much appreciated you digging into the single-CODEC/stereo side of it.

I'm happy with treating the full per-VFO work as a separate MR after this one lands. The L/R stereo mapping (MAIN on L, SUB on R, mapped to A/B) plus the bench channel-map check is clearly its own piece of work, and there's no sense holding the whole audio feature up for it.

The one thing I'd like to pull forward into this PR, if it's a small change, is your step 1 - stop attaching the live FFT to VFO B and drop it back to the decorative fallback while the audio is mono. My reasoning is just that I'd rather not ship a VFO B trace that looks live but is actually the mono/MAIN signal - on a dual-receiver radio that's the kind of thing a user will read as real and trust, so I'd sooner it be obviously decorative until the stereo mapping is in. If that turns out to be more than a quick change, tell me and we'll merge as-is and do it all in the follow-up.

The blank/non-radio TX device guard as a follow-up sounds spot on - that was the whole cause of the ringing, so it'll save the next person the same head-scratching.

Once step 1 is in (or you confirm it's not a quick one), I'm ready to merge. Thanks again for all of this.

Colin

fabiojvalente:
Hi Colin,

Step 1 is implemented, the VFO B for now uses the Math.random() for the scope

**As for the TX/RX Guard**

I’ve tightened that without hard-filtering device names (operators often rename the endpoint to e.g. “FTDX101”, and OS strings vary):

1. **Blank TX/RX guard** — when Remote Audio is enabled, Settings save requires explicit RX and TX devices; the bridge no longer opens the OS defaults.
2. **Soft hint only** — names containing USB Audio / Yaesu sort first and get a 📻; the full list stays available for renamed devices.
3. **Clearer Settings + manual** — RX = radio recording / Microphone|Line (USB…); TX = radio Speakers/playback — not PC speakers; troubleshooting for “I hear myself in the speakers”.

mm5agm:
Merged to develop - thank you, Fabio, this is a genuinely significant piece of work.

Step 1 is exactly what I wanted: with VFO B on the decorative fallback there's no risk of anyone reading that trace as a real second receiver while the audio is mono, so I'm happy to ship it like that until the stereo mapping lands.

And thank you for pulling the blank TX/RX guard forward into this PR rather than leaving it for later - that was the whole cause of the ringing on my bench, so requiring an explicit RX and TX device (instead of falling back to the OS defaults) will save the next person the same head-scratching. I like that you did it by sorting/hinting rather than hard-filtering on device names too - operators rename the endpoint all the time, so a soft radio hint that still shows the full list is the right call.

I've merged it as-is. It's on develop, not in a public release yet - I'll fold it into the next release once it's had a bit more mileage, and I'll make sure the release notes credit you.

That leaves the per-VFO stereo scope (MAIN/SUB on L/R, mapped L->A / R->B on dual-receiver radios) as the separate follow-up MR we agreed, whenever you're ready to pick it up - and I'll do the bench channel-map check on the 101 for that since it needs real hardware.

Thanks again for all of this.

Colin

Issue #48, opened 2026-06-18
[FEATURE] Re-arrangement of the controls in YWC GUI


SP3L-Jacek:
I often loose some time when I search for the control I would like to change. I can not remember in which part of the window I can find it. At the beginning, I thought it was a matter of getting used to the YWC layout. But time passes and it is not getting any easier for me.

So, I thought how I would arrange the controls if I were responsible for the YWC GUI. I would arrange the controls in six groups as in the attached picture. Such an arrangement would be intuitive for me. Having said that I realize that this is only my personal view. So, consider it as an idea for further discussion.

<img width="524" height="1028" alt="Image" src="https://github.com/user-attachments/assets/d7ea4228-11fe-4953-9674-8d0cdbe6a98a" />

mm5agm:
Thanks Jacek — really appreciate you putting the layout sketch together. The current layout reflects a lot of feedback from other testers and a wholesale rearrange would break their muscle memory, so a full redesign isn't something I want to take on.

But your underlying point — that you lose time hunting for controls — is one I want to address if I can. Could you tell me which **two or three specific controls** you spend most time looking for? If they have a common pattern (e.g. all the noise-reduction-related controls are scattered across the page, or all the "less common" controls are at the top instead of grouped at the bottom) we could improve findability for *those specific things* without a full redesign.

Leaving this issue open as a parking lot for the specific findability improvements that come out of that.

73, Colin

SP3L-Jacek:
OK, I understand. Because I am currently focused on testing rather than using it for making QSOs, I can not say which controls are the most difficult in normal life. I will choose 2 or three most difficult to find when I move on to normal usage of the application.

SP3L-Jacek:
Well, I couldn't resist and designed my own GUI for YWC.
Maybe, at some moment in the future, it will become possible to make a revolution in the present GUI?

<img width="1790" height="863" alt="Image" src="https://github.com/user-attachments/assets/b412df12-6a54-4c44-91e5-4907c817c8ff" />

mm5agm:
Jacek — I like the mockup, that's a really clean dashboard. Single horizontal meter strip across the top, both VFOs side by side with their controls densely packed — it would feel very natural for someone running a single radio at a focused station.

Two questions before I think any harder about implementation:

[...]

PR #90, opened 2026-08-05
Add macOS / Linux CAT-only host, Docker (amd64/arm64)

fabiojvalente:
⚠️ This is still work in progress and not ready for merging

Tests are welcome

## Summary

Makes Yaesu Web Control runnable as a **CAT + web UI** host outside Windows, without shipping SDR spectrum or Voice Control on those platforms. The Windows installer path (`net10.0-windows`) is unchanged.

- **Multi-TFM:** `net10.0-windows` (full product: WinForms tray, voice, SDR worker) and `net10.0` (CAT-only console host for macOS/Linux).
- **macOS menu-bar tray** via Avalonia (`MacSystemTrayService`), driven on the process main thread after Kestrel `StartAsync` (AppKit-safe). Linux uses plain `app.Run()` with no tray.
- **`AutoShutdownWhenNoBrowsers`** setting (default on) so a headless shack can keep the process alive with no browser tabs; **Docker forces keep-alive** and skips local browser open (`HostRuntime.IsContainer`).
- **Linux Docker** multi-arch image (`linux/amd64`, `linux/arm64`) + `docker-compose.yml` with serial device pass-through and a data volume under `XDG_CONFIG_HOME=/data`.
- **UI/docs:** Settings/Index gate Windows-only panels with `IsWindowsHost`; serial validation accepts `/dev/…` on Unix; README / USER_MANUAL / CLAUDE document the operational differences (platforms table, install paths, FAQ §15.10).

### Notable code

| Area | Change |
|------|--------|
| `Yaesu_Web_Control.csproj` | Dual TFM; compile-remove WinForms tray / voice / SDR controller on `net10.0`; Avalonia + DBus pin on CAT-only |
| `Program.cs` | `#if WINDOWS` for tray/voice/SDR; macOS tray main-loop; Linux `app.Run()` |
| `RadioHub.cs` | Respects auto-shutdown setting; containers never auto-exit |
| `Services/MacSystemTrayService.cs` | Open / About / user data / Exit |
| `Dockerfile`, `docker-compose.yml`, `docker/entrypoint.sh` | Publish `net10.0` RID image; data dir under `/data`; container defaults via `HostRuntime` / `ApplyContainerDefaults` (no baked settings file) |
| CI `release.yml` | Explicit `-f net10.0-windows` for the Windows publish |

## Test plan

- [x] `dotnet build -f net10.0-windows` succeeds (Windows product TFM; OK with `EnableWindowsTargeting` on macOS)
[...]

PR #93, opened 2026-08-06
Feature/rnd UI experimentation

fabiojvalente:
This is simply some RND , and is very far from replacing the current ui (plenty of functionality missing)

Opening the MR to gather feedback and have a open discussion while researching potential (or if even needed/desierable) for a new ui
https://github.com/mm5agm/Yaesu_Web_Control/issues/48

This has a bit of a personal opinion in it and as such some of the goals are:

- Be easy to use without overloading the user with too much information
- Cover usage on Desktop, Tablet and Mobile
- Keyboard shortcuts for controls
  - Switching between VFOs
  - Switch mode  
  - Set frequencies
  - MOX / PTT
  - open any subpanel (ex clarifier, dsp, cw settings, filters)

Current state

**Desktop view**
Screenshot 2026-08-06 at 14 12 57
image

**Tablet view**
image
image

**mobile view**
image
image


mm5agm:
**Fabio — read the first part before you write any more code**

Thank you for this. It's the third substantial thing you've brought me and the second that solves a problem I hadn't got to yet. But there's something you need to know today rather than in a fortnight, so I'm putting it first.

## The bit that's time-sensitive

**I'm changing the way alternative layouts work, and it affects what you've built.**

You couldn't have known this, but there's a parked Phase 7 in `Plan.md` covering exactly this ground — Jacek SP3L's alternative layout, from the same issue #48 you linked. It's been blocked since June on one unanswered question: *"where would the SDR spectrum displays go?"* Jacek never came back on it. **Your screenshots answer it.** You've unblocked a phase you didn't know existed.

The design I agreed with Jacek was a second Razor page plus a companion CSS file, with a "GUI Layout" dropdown in Settings choosing between them. That is almost exactly what you've built — `UiV2.cshtml` + `ui-v2.css`. So to be clear: **you didn't go the wrong way. You went the way I'd already committed to.** There's even a `feature/jacek-gui` branch with the Settings dropdown skeleton on it, which you should look at before building another one.

I'm changing it anyway, and the reason is arithmetic. Two layouts is fine. Three — Classic, SP3L, yours — means every new control gets added three times, every bug fixed three times, every accessibility label written three times, and they drift apart within a couple of releases. I'm one person and I won't keep up with that. So:

**Layouts become skins. One set of markup, shared by all of them. A skin is CSS.** Every control group becomes a named grid area, declared once; each skin is a stylesheet supplying its own `grid-template-areas` and hiding the groups it doesn't want. Adding a control costs the markup once plus a line in each skin's map. Fixing a bug costs once.

**What that means for `UiV2.cshtml`, honestly: most of the markup goes.** Not the thinking — the arrangement, the panel grouping, the decisions about what to promote and what to bury, the tablet and mobile work. That's the valuable half and it survives intact. But the 1,318 lines of HTML get re-expressed as area names and a stylesheet rather than kept as a page.

I'd rather tell you that now than let you spend three more evenings on it. And I want to be straight about the cost to you, too: CSS can move a control anywhere on the page, but it can't move it into a different parent. If part of your design needs a control nested somewhere the shared markup doesn't put it, the skin can't express it. My maintenance burden goes down and your layout freedom goes down. That's the trade I'm making, and you should know I'm making it at your expense as well as mine.

## What I'd like from you instead — three things, all of which survive

**1. Keyboard shortcuts.** From your goals list, and I'd take them on their own merits, into the current UI, right now. They're layout-independent, so nothing you write gets thrown away later. A keyboard-only path is also a screen-reader path, which matters to me more than most things (see below).

**2. The control manifest.** Not a list of controls — your screenshots answer that better than a list would. One level up: **which control *groups* your layout shows, which it collapses behind a sub-panel, which it drops entirely, and why.** A page of prose. That becomes the direct input to the skin design, and if the mechanism I'm about to build can't express what you want, I'd much rather find out from your document than after I've written it.

**3. Testing #90 on a real radio.** See the offer at the end.

You should also talk to Jacek on #48. You're both solving the same problem from different ends, and his layout and yours will be the first two skins.

## Responsive layout — a straight question

Your desktop/tablet/mobile work is the part I'm least sure the skin mechanism handles well, and you've thought about it harder than I have. Is responsive behaviour a *fourth* skin, or is it something every skin has to carry via media queries? If you have a view, I'd genuinely rather build to it than guess.

## Accessibility — a condition, and I'd rather explain than assert

Voice control and screen-reader labelling exist in these apps for partially-sighted operators. It's the reason the Icom sibling exists at all. If a skin ships in my installer and an operator picks it and loses voice control, that's a regression in *my* app — they won't know a contributor wrote it, and they'd be right not to care.

The good news is it costs you very little:

- The voice intents talk to **HTTP endpoints, not the DOM**. `IntentDispatcher` doesn't care what the page looks like — most of voice control works in any skin for free.
- Screen-reader labels bind on a **`data-a11y-key`** attribute. That's an attribute on a button, not a design constraint.
- My current UI is not the standard to beat here. The Labels page advertises 28 keys that no element actually carries — dead rows I haven't got round to fixing.

And something that may interest you rather than burden you: your **tablet** layout — big hit targets, less clutter, larger readout — is close to a skin I already wanted, aimed squarely at partially-sighted operators. You may have accidentally built it.

The line: an experimental skin behind a flag, unlisted in Settings, can ship without full accessibility support and be labelled as such. One that's offered in the dropdown or made default can't.

## On the diff

For anyone else reading: #93 shows 57 files and 6,600 additions because it's branched on top of #90 and #92. The genuinely new work is the final commit. Reviewing #93 as it stands means reviewing all three at once.

## On #90 — a straight question

The code isn't what worries me; the commitment is. Merging it makes Linux and macOS platforms I nominally support and **cannot test**, on hardware I don't own. Would you be willing to be named maintainer for those platforms — documented as community-supported, CAT-only, with breakages coming to you? If yes, I'm inclined to take it. If not, I'd rather not claim a platform at all, because half-supporting one is worse for users than being honest that it isn't supported.

## The offer

You've been careful about the limits of your testing — *"only tested on the FT-DX-10"*, *"not sure what's the behaviour on dual receiver units"*. You're right to be, and you can't test what you don't have. I can fix that.

You said the Pi/Linux host is the first stage of the remote setup you actually want, now that the SCU-LAN10 has turned up without a PSU. **A Pi in the shack, wired to the radio, serving the web UI is your target architecture.** So rather than me testing your arm64 image badly on a platform I don't use — let's put it on my radio and let you drive it.

What I'm proposing:

- I put 64-bit Raspberry Pi OS on the 3B, wire it to the **FTdx101MP** — dual receiver, the case you've never been able to test — and run your Docker image on it.
- **The Pi gets its own CAT port.** The FTdx101MP has two independent CAT interfaces, USB and RS-232, and they run simultaneously — same command set, neither blocks the other. So the Pi takes one and my PC keeps the other, and we're not taking turns or unplugging anything. Both can be live at once, which also means you'll see my changes appear on your screen and vice versa.
- **Tailscale** between it and your Mac. You get a private address, you log into the Pi as if you were in the room. Nothing exposed to the internet, which matters because the app has **no authentication whatsoever** and anything that can reach port 8080 can key my transmitter.
- It's your box for the duration. Root, reflash it, break it — there's nothing of mine on it.

Two limits, both honest:

**It's CAT-only.** So you get the dual receiver, but not the SDRplay spectrum and waterfall you've never seen running, and not voice control. Both are Windows-only and your own #90 fences them off correctly. If you ever need those, that's a different conversation about my actual PC, and we can have it then.

**Receive only, and enforced in software.** I'm licensed for everything that leaves this station. We're under the same regulator, which makes it simpler than it might be, but I still want to read the terms on remote and third-party operation properly rather than assume. The practical problem is that I have no 200 W dummy load, so "unplug the antenna" isn't a safe answer with a radio that can be keyed by CAT. **So I'll build the inhibit: an `AllowTransmit` setting, off for your sessions, enforced at the CAT-send layer rather than in the browser** — that part matters, because rigctld and WSJT-X can key the radio without the web UI being involved at all, and a checkbox on the page wouldn't stop them.

That setting is worth having regardless — for demos, for screenshots, and for anyone who wants to let a visitor drive the receiver.

Only loose end is which end takes which port. Whichever the Pi ends up on it'll want a `YWC_SERIAL_DEVICE` path in the compose file and 38400 baud, and if it lands on the RS-232 side it needs a USB-serial adapter, since the 3B has no DB-9. I'll sort that out this end.

## Last thing, going the other way

There's currently no way to run YWC at all without a radio — no simulator, no stub. The Icom sibling has one and it's the difference between a contributor being productive on their first evening and giving up. I'm going to build the equivalent here: a simulated CAT client behind `ICatClient`, answering `FA;`, `MD0;`, `SM0;`, `RF;` and friends from an in-memory state, plus a synthetic spectrum feed so the SDR panels light up. Layout work with no radio, on any platform.

Fair warning when it lands: a simulator lies about timing. UI built against it can look perfect and still fall over on a real 38400-baud bus. It's for layout, not protocol.

73, Colin MM5AGM

fabiojvalente:
Hi Colin,

To be clear, this MR was never meant to be merged, but instead share some concepts, opening the MR was just to track the progress if any (hence the whole separate stack for the UI, to leave the existing one as is/unaffected) :)

I will reach out via email regarding the other points if that's okay 

I'll be closing this one for now to prevent any further confusion

PR #97, opened 2026-08-12
Feature/add video tx

### LINKED ISSUE #96: [FEATURE] Display radio video out in browser
## What would you like YWC to do?

Recent Yaesu radios provide a DVI-D video output intended for an external monitor. A USB video-capture device can be connected to this output, allowing the radio’s front-panel display to be digitised and made available to the browser.

_At this stage, this is primarily an expression-of-interest feature request rather than a fully specified implementation proposal. I am mainly trying to determine whether there is sufficient interest from other users and contributors in having the radio’s external display available inside Yaesu Web Control._ 

## Why is this useful?

This would complement the existing CAT-based remote-control functionality and provide visual feedback when operating the radio remotely.

Yaesu Web Control already provides browser-based transceiver control and real-time spectrum/waterfall functionality when an SDR is available. The requested feature would add the radio’s actual display as an optional browser view, specially in scenarios where a SDR does not exist.

### Low-cost implementation
The hardware required for this feature is relatively inexpensive and widely available. A typical setup would only require:

- A DVI-D-to-HDMI cable compatible with the radio’s video output.
- An HDMI-to-USB video-capture device connected to the computer running Yaesu Web Control.

## Are there any references?

- Yaesu DVI-D to HDMI. Investigation - https://www.youtube.com/watch?v=dtlziYEsXxE
- Capturing yaesu video - https://www.youtube.com/watch?v=1EY7m5e91TI

## Which radio(s) and operating mode(s) would benefit?

- FT-710
- FTDX10
- FTDX101D
- FTDX101MP

Notes: Am wondering if this is something that wou

fabiojvalente:
## Summary
- Adds **Radio Display**: capture from a USB webcam or HDMI capture dongle and stream it to the browser as MJPEG (`/api/video/stream`).
- Capture opens on the first viewer and releases when the last viewer disconnects. Device, FPS (15 / 30 / 40 / 60), max width, and JPEG quality persist in settings; device and FPS can also be changed from the Index panel.
- Index card plus a dedicated `/RadioDisplay` page (fit, fullscreen, pop-out). Device enumeration covers Windows/macOS and Linux (`/sys/class/video4linux`); Docker maps `/dev/video*` and the video group.

## Test plan
- [x] Enable Radio Display in Settings, pick a capture device, confirm Index panel and `/RadioDisplay` show a live stream.
- [x] Change device and FPS from the Index controls; confirm they persist and the stream restarts.
- [x] Fit / fullscreen / pop-out behave as expected; stream stops when all viewers close.
- [x] Disable Radio Display in Settings → stream and device APIs return 403.
- [x] Linux/Docker: `/dev/video*` mapped and video GID set; devices list and stream work.
- [x] No capture device / unplug while streaming → status/error is shown without crashing the host.


mm5agm:
Tested on Windows 11, .NET 10, with an MS2109 HDMI capture dongle (`VID_534D&PID_2109`) on an FTdx101MP. Branch at `50bb40e`.

**The good news first:** once running, it works well. Streaming held **800×600 at 15.0–15.1 fps** against a 15 fps target, rock steady across repeated samples — the dongle honoured the requested size exactly, so no downscale was needed. The radio's screen is perfectly readable in the browser. Unplugging the dongle correctly emptied the device list, set `status: disconnected` with a sensible error, and left CAT control completely unaffected — I could still change bands with video down.

But there are five Windows-specific problems, and I think three of them are structurally invisible on macOS.

---

## 1. Critical — reloading the page while streaming crashes the whole process

Repro: start streaming, then press **Ctrl+F5**.

```
Fatal error. 0xC0000005
   at OpenCvSharp.Internal.NativeMethods.videoio_VideoCapture_delete(IntPtr)
   at OpenCvSharp.VideoCapture.DisposeUnmanaged()
   at OpenCvSharp.DisposableObject.Dispose()
   at Yaesu_Web_Control.Services.Video.VideoDeviceEnumerator.ProbeIndices(...)
   at Yaesu_Web_Control.Services.Video.VideoDeviceEnumerator.ListDevices()
   at Yaesu_Web_Control.Controllers.VideoController.ListDevices()
   ...
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol+<ProcessRequests>
```

Process exit code 5.

The page load asks for the device list. `ProbeIndices` opens and disposes a `VideoCapture` on every index 0–15, including the index the capture loop is streaming from right then. Disposing that contended DirectShow handle faults in native OpenCV.

The reason this is critical rather than cosmetic: **0xC0000005 is a native access violation, so the `try`/`catch` in `ProbeIndices` cannot catch it.** The CLR fails fast and the entire host goes with it — CAT, SignalR, the tray app, everything. The video subsystem is otherwise nicely isolated (see the unplug result above), but this bypasses that isolation completely.

Suggested fix: never probe an index that the capture service currently has open. Either share a lock between `VideoDeviceEnumerator` and `VideoCaptureService`, or have the enumerator skip the active index and have `VideoController` splice that entry in from the capture service's own known-good state.

## 2. Blocker — the branch does not compile on Windows

```
Services\Video\VideoCaptureService.cs(257,60): error CS0104:
'Size' is an ambiguous reference between 'System.Drawing.Size' and 'OpenCvSharp.Size'
```

YWC's `net10.0-windows` target sets `UseWindowsForms=true`, and WinForms projects get an implicit `global using System.Drawing`, which collides with `OpenCvSharp.Size`. You won't see this on macOS because that global using isn't there.

One-line fix, safe on all platforms — add to the top of `VideoCaptureService.cs`:

```csharp
using Size = OpenCvSharp.Size;
```

I applied this locally to do the rest of the testing.

This one argues for adding a **Windows CI build** to the workflow before merge — it would have caught this on its own, and the project ships primarily to Windows users.

## 3. Bug — capture loop wedges at `connecting` after a replug

After unplugging the dongle for a few minutes and plugging it back in, the panel stuck on `Connecting`, showing `Index:0 (not present)`, and never recovered on its own.

Evidence it was the app and not the hardware:

- Capture retries (three backend attempts each, every 2 s) ran from t=604 s to **t=803 s, then stopped permanently**.
- A later enumeration logged **15 warnings across 16 probes** — one index opened silently, i.e. the dongle was available and openable the whole time.
- Status sat at `connecting`, `error: null`, `viewers: 1`, with no open attempts being made at all.
- `POST /api/video/device` (i.e. reselecting the device in the panel) recovered it to `streaming` in under 4 seconds.

So the device was fine and the loop simply wasn't trying. A shorter unplug/replug *did* self-heal, so the trigger isn't the unplug duration.

I'm not certain of the mechanism. One candidate is the viewer-count gate: `CaptureLoopAsync` sets `connecting` and then gates on `_sessions.ViewerCount > 0` (line 184), while `StartLoop` early-returns if the previous task hasn't completed (lines 150–151) — a page reload momentarily drops viewers to zero, which could leave the status set but no loop running. But the same device contention behind finding 1 could equally explain it, so please treat that as a hint rather than a diagnosis.

Workaround for anyone testing: reselect the device in the panel dropdown.

## 4. Minor — no friendly device names on Windows

Every device shows as `Camera 0`, `Camera 1`, … because `ListDevices` passes `ReadEmptyNames()` on the Windows path, while Linux reads v4l2 names and macOS shells out to ffmpeg. With a webcam and a capture dongle both plugged in there's no way to tell which is the radio.

Windows can supply the name — this returns `USB Video` for my dongle:

```powershell
Get-PnpDevice -Class Camera -Status OK | Select-Object FriendlyName, InstanceId
```

The catch is mapping a PnP entry to an OpenCV index; DirectShow enumeration order usually matches, but it isn't guaranteed. Since the label already carries `(#N)`, a best-effort name plus the index would still be a clear improvement over `Camera 0`.

## 5. Nit — stale numbers on the disconnected banner

After unplug the panel reads `Disconnected: 800x600 @ 15.1fps` in white-on-red. `/api/video/status` keeps `width`, `height` and `fps` at their last-good values while `status` is `disconnected`, so it reads as though it's still streaming. Zeroing them on disconnect (as `StopLoop` already does for the frame buffer) would match what `width: 0` already does in the `connecting` state.

---

Happy to open a PR against your branch with the fix for 2, and to retest 1 and 3 once you've had a look. Let me know which way you'd rather take the enumerator fix and I'll try it here against real hardware.

fabiojvalente:
@mm5agm what commit did you test on ?

What you describe should've been fixed on the latest one

mm5agm:
Correction to one detail above, since it would give you a misleading picture of the size negotiation.

I said the dongle "honoured the requested size exactly" at 800×600. That was true of the first session, but not in general. After restarting the app and replugging the dongle into a different USB port, the same `cap.Set(FrameWidth, 800)` / `cap.Set(FrameHeight, 600)` request came back as **720×576**:

```json
{"status":"streaming","width":720,"height":576,"fps":15.1,"viewers":2,
 "maxWidth":800,"targetFps":15,"jpegQuality":65}
```

Same dongle, same settings, same code — just a different USB port and a fresh process. So on Windows/DirectShow the size really is best-effort, as your comment on that line says, and it can land on a different format between sessions. Both 800×600 and 720×576 were perfectly readable for the radio's screen, and frame rate held at 15.1 fps either way, so this isn't a problem in itself — I just don't want to leave you thinking 800×600 is reliably negotiated when it isn't.

One incidental positive while I'm here: after the crash in finding 1, restarting the app resumed streaming automatically from the persisted `VideoDisplayEnabled` / `VideoCaptureDeviceKey` settings as soon as a browser reattached. No reconfiguration needed.

fabiojvalente:
@mm5agm please check the latest commit as I forgot to push 🤦‍♂️ 

`36f0d4c6b4cf09849c1662acd92f05380b0ae6dd`

mm5agm:
First, a correction on my side: I tested **`50bb40e`**, and `36f0d4c` (11 Aug) is already an ancestor of it — so my report was against code that included the commit you pointed me at. What I *hadn't* got were the two you pushed at 14:24 today, `90a43d5` and `6812e07`. I've now retested on `6812e07`.

Two are fixed, one has changed shape into something I think is worse, and two are still open.

## ✅ 1. Crash on device enumeration — fixed, verified

`ListDevices(allowProbe: !_capture.IsCapturing)` plus the `OpenDeviceIndex` fallback does the job. I hit `GET /api/video/devices` repeatedly while streaming — the exact call that killed the process before — and the host stayed up with the stream uninterrupted. I also re-ran the original user-level repro, **Ctrl+F5 while streaming**, and it no longer crashes. Nice fix, and `WindowsDshowDevices` binding to property bags without opening the graph is the right way round.

## ✅ 2. Windows build break — fixed, verified

`new OpenCvSharp.Size(...)` resolves it. Builds clean on `net10.0-windows` with my local patch removed. I'd still suggest a Windows CI job so this class of thing can't reach a reviewer again.

## ❌ 3. Regression — unplug is no longer detected at all

This is the one I'd look at first. At `50bb40e` an unplug was reported within seconds as `disconnected` with a sensible error. At `6812e07` it isn't detected at all:

```
15:42:48  status=streaming  800x600 fps=1  err=
15:43:20  status=streaming  800x600 fps=1  err=
15:44:00  status=streaming  800x600 fps=1  err=
```

Over a minute with the dongle physically unplugged, still reporting `streaming`, `error: null`. The panel showed no picture the whole time — so the app reports healthy while displaying nothing, which is harder for a user to act on than the old honest `disconnected`.

Cause looks like the empty-frame heuristic:

```csharp
if (!cap.Read(frame) || frame.Empty()) { emptyStreak++; ... }
emptyStreak = 0;
```

`cap.Read()` keeps returning success with a non-empty frame after the device is gone — DirectShow hands back stale frames rather than failing — so `emptyStreak` resets every pass and `EmptyFrameGiveUp` never trips. The `fps=1` reading fits: frames still "arrive", just slowly.

Why it used to work: the error at `50bb40e` was `"Could not open capture device index 0."`, which comes from the *open* path, not the empty-frame path. That loop broke out and failed on reopen; this one never breaks out.

**It also doesn't recover when you plug it back in.** 30 s after replugging it was still `fps=1`, still `streaming`, still no picture — because the loop never exits, it never reopens. `POST /api/video/device` (reselecting in the panel) recovers it to a healthy 800×600 in about 6 s, so that's still the workaround.

`frame.Empty()` may just be the wrong detector for UVC/DirectShow. A wall-clock watchdog on the last *changed* frame, or treating a sustained collapse in measured FPS against target as a disconnect, would catch what this misses.

## ❌ 4. Friendly names — code is there, but returns nothing here

`WindowsDshowDevices.ListFriendlyNames()` appears to come back empty on my machine: the dropdown still shows `Camera 0  640×480`, and the `allowProbe: false` path falls through to `_cache` rather than the named list.

The name is definitely available from Windows — for the same dongle:

```
FriendlyName : USB Video
InstanceId   : USB\VID_534D&PID_2109&MI_00\7&35A07035&0&0000
```

So it looks like the COM enumeration isn't returning monikers rather than the names being absent. Possibly apartment state on the request thread, or `CreateClassEnumerator` returning `S_FALSE` (1) — worth noting the `hr != 0` check treats `S_FALSE` as failure, which is correct for "no devices" but would also mask a category mismatch. Happy to add logging and tell you exactly what it returns here if that helps.

## ❌ 5. Stale numbers on the disconnected banner — still open

The three `SetStatus("disconnected")` sites still don't clear `_frameWidth`/`_frameHeight`. Largely academic while finding 3 stands, since it never reaches `disconnected` on an unplug anyway.

## ❌ 6. New — the panel never reconnects the stream after an interruption

Once the capture loop had been recovered and the server was demonstrably streaming again, the browser panel still showed nothing. That turned out not to be a server problem at all — pulling `/api/video/stream` directly gave **5.4 MB across 313 chunks in 6 seconds**, with sampled chunks distinct, so real changing frames were going out the whole time.

The panel's `` MJPEG connection dies with the capture interruption and is never re-established. The server restarts the stream underneath it, but the client doesn't reconnect and there's nothing in the UI to say so — you're left looking at a blank panel while `/api/video/status` says `streaming`. A hard reload brings the picture straight back.

Worth an `onerror`/`onstalled` handler on the image that re-attaches the stream (with a backoff), so a transient capture problem doesn't need a manual page reload to clear.

---

## Minor observation

Throughput is a little down: `50bb40e` held 15.0–15.1 fps against a 15 target, `6812e07` sits at 12.8. Same dongle, same settings, same 800×600. Not a problem at these rates and it may well be measurement noise across a restart, but flagging it in case the stability changes cost something in the read loop.

Everything else from the first pass still holds — 800×600 is readable, CAT control is unaffected by video trouble, and two browser sessions share a single capture loop correctly.

fabiojvalente:
Thanks @mm5agm — this was a really useful report. Branch has been rebased onto current `develop`, so the SHAs you cited have moved; the behaviour below is on the current tip (`e3a6c70`). The crash and the Windows build were already in good shape on your pass. 3–6 have been addressed since the commit you tested (`c4c05ba`, was `6812e07`). Could you re-run those on this build?

### ✅ 1. Crash on device enumeration
No change since your verification. `ListDevices(allowProbe: !_capture.IsCapturing)` plus the `OpenDeviceIndex` fallback stay as-is. Agreed on property-bag enumeration without opening the graph.

### ✅ 2. Windows build break
Still good — `new OpenCvSharp.Size(...)` is in. Your CI suggestion is already on `develop` after the rebase: PRs into `develop` now run `build-windows` (`dotnet publish -f net10.0-windows`), from #101. This PR will pick that up once pushed.

### ✅ 3. Unplug no longer detected
You were right: `cap.Read()` keeps succeeding with a stale non-empty frame, so `frame.Empty()` never trips and the loop never exits.

`frame.Empty()` is still there as a first check, but unplug is now detected the way you suggested (`c9d8c63`):

- **Slow-read watchdog** — if `Read` blocks ≥ 750 ms for 4 frames in a row, treat as unplug (`Capture device stalled (no timely frames).`).
- **FPS collapse** — after a 2.5 s warmup, if measured FPS stays below 25 % of target for 3 one-second windows, same outcome (`Capture device stalled (frame rate collapsed).`). That is the `fps=1` case you saw.

Both paths call `MarkDisconnected`, **break out of the capture loop**, release the graph, and reopen after a short settle. Replug should recover without `POST /api/video/device`. The USER_MANUAL troubleshooting row now says unplug is reported as **Disconnected**, then the host reopens on its own.

Please re-test the exact sequence: stream → unplug for ~30 s → confirm `disconnected` (not `streaming` / `fps=1`) → replug → picture comes back without reselecting the device.

### ✅ 4. Friendly names
Your apartment-state guess was the cause. DirectShow’s system device enumerator is STA; the `/api/video/devices` request thread is MTA, so `CreateClassEnumerator` was returning no monikers and we fell through to `Camera 0`.

`WindowsDshowDevices.ListFriendlyNames()` now hops to a dedicated STA thread when the caller isn’t already STA. The `allowProbe: false` path uses that named list instead of `_cache` when COM returns anything.

Expected dropdown label for your dongle is **USB Video**, not `Camera 0  640×480`.

### ✅ 5. Stale numbers on the disconnected banner
There is now a single `SetStatus("disconnected")` site, `MarkDisconnected`, and it zeros `_frameWidth` / `_frameHeight` (and measured FPS) first. Academic until 3 fires, as you said; it should come along for free once unplug is actually detected.

### ✅ 6. Panel never reconnects after an interruption
Agreed it was a client problem. The `` now has `onerror` / `onstalled` / `onabort` and re-attaches `/api/video/stream?t=…` with backoff (2.5 s, capped at 15 s).

MJPEG `` can also die quietly without firing those, so the 4 s status poll also re-attaches when:

- the server goes back to `streaming`,
- `frameSeq` resets (capture loop restarted),
- or the image has been blank for two polls while the badge says streaming.

A hard reload should no longer be required. The manual has a row for “panel blank while badge says Streaming”.

---

**Please re-verify 3, 4, and 6** on this build — those were the ones that needed a real HDMI dongle. 1 and 2 should be unchanged from your last pass.

Finally, the don’t treat a ~20 fps reading as a regression of 3–6 unless the badge is stuck on streaming after an unplug. 

**The usb hdmi capture needs to run on usb 3** to ensure the 30fps are met**

mm5agm:
I was wondering whether the 800x600 display could be "upscaled" to show bigger and sharper, so I went digging through how the stream scales. Sharing what I found in case it's useful - no changes needed from me, it's your branch and your call.

**The short version for the upscaling question:** true upscaling (interpolation or AI) can't add detail the radio's DVI output never sent, and AI super-res on a live stream would be far too heavy anyway - plus it risks distorting the frequency digits, which we can't have. So I steered away from that. But a couple of things stood out that are genuinely worth knowing:

- The 800x600 isn't a hard limit from the radio - it's the `VideoMaxWidth = 800` default cap plus the requested capture size. Raising it only helps if the dongle/DVI actually emit more than 800 wide, so it's worth testing what native resolution your dongle reports.
- `VideoJpegQuality` defaults to 65 (max 85). A lot of the softness people notice when they enlarge the image is JPEG compression rather than resolution, so bumping quality is probably the cheapest real "crispness" win.
- Both of those persist in settings, but they're not surfaced in the UI yet - only the on/off toggle is on the Settings page, and device/FPS are on the Index panel. If you wanted, a quality/width control next to the FPS picker would let people tune it live instead of editing JSON. Totally optional, and entirely your call.

One tiny thing I noticed while reading: the capture requests `FrameHeight = 600` regardless of `VideoMaxWidth`, so raising max width alone keeps the requested height fixed. Might not matter in practice if the dongle ignores the requested size, just flagging it.

**One longer-term thought this sparked** - and very much your call since it touches your design: I'd played with the idea of putting clickable controls over the video so you could action the radio from the streamed screen. Two ways to do it, and I think only one is worth it:

- Invisible hotspots pinned directly onto the video - I'd rule this out. The radio draws Span/Fix/3DSS as soft-keys inside its own TFT, so they move with menu and mode, and hotspots tied to fixed video pixels would drift. There's no "tap at x,y" CAT command to fall back on.
- A "virtual front panel" instead - the video fills one fixed rectangle, and YWC's own buttons sit at stable positions around it, wired to the CAT commands (Span, Hold, Fix, 3DSS are all reachable via `SS`; Expand isn't CAT-addressable at all). This way we never chase the radio's moving labels, and video alignment actually gets easier.

The honest caveat is that this would re-couple the deliberately-CAT-free video page back to CAT, so it's genuinely your decision whether that fits where you want this feature to go. Not asking for anything here - just parking the thought next to the code it relates to.

fabiojvalente:
Linux tested
A Raspberrypi is now capable of streaming at 15fps (usb2) - though not recommended user experience might be affected as this eats all its cpu and audio becomes choppy :)

Pending another branch containing the required fixes for macos

fabiojvalente:
@mm5agm this has now been tested across the 3 OSs, I will now work on the cosmetic/ux/ui bits

On the three points:

- **Native resolution** — what the radio outputs is what the dongle is fed (typically 800×480 or 800×600). That is not automatically what the host sees. Most HDMI capture sticks don’t expose an 800-wide UVC mode; they scale or letterbox into 720p / 1080p (or another pin they advertise). Which pin you get is hardware-dependent, so we can’t guarantee a single negotiated size. Raising VideoMaxWidth only avoids downscaling that dongle frame; it doesn’t recover pixels the radio never sent.

- **JPEG quality** — this is the one worth chasing. Default 65 is a likely source of the softness when the image is enlarged. I’ll bump quality and expose it in the UI. Caveat: on Windows MJPEG passthrough the dongle’s JPEG is forwarded as-is, so VideoJpegQuality only applies on the decode-and-re-encode path.

- **UI** — I’ll add quality (and max-width) next to the FPS picker on the Radio Display panel in this PR, same persist-and-live-apply pattern as FPS. Default max-width stays 800 unless a given dongle is truly capturing the panel wider than that.

**On the height note:** current code requests paired 4:3 modes (800×600, then 1024×768 if the cap is raised), not a floating width with height stuck at 600. Encode downscale already keeps aspect. Many dongles ignore the requested size anyway, and the Windows MJPEG path doesn’t use that table. For an 800×480 radio that usually means the panel sits inside a 4:3 or 16:9 capture (letterbox), not cropped digits. Impact looks low in practice — I’ll keep an eye out for a stretched 800×480 panel, but I won’t treat it as a blocker.

Will ping again once jpeg quality and the ux as been updated :)

mm5agm:
Retested on `2110473`. Findings 1, 3, 4, 5 and 6 all pass. Details and timings below, plus two observations and one bug I hit on the way.

**Environment**

- Windows 11, FTdx101MP front panel via MS2109 HDMI capture
- Capture path: OpenCV DirectShow (`passthrough=false`, YUY2 800x600, encode mode)
- Second capture device present throughout (OBS Virtual Camera at index 1)

---

**Finding 4 - friendly device names: PASS**

`GET /api/video/devices` returns real names for both devices, not `Camera 0` / `Camera 1`:

```json
{"devices":[{"key":"index:0","label":"USB Video","index":0},
            {"key":"index:1","label":"OBS Virtual Camera","index":1}]}
```

Checked both idle and while streaming, so the `allowProbe: false` path is covered. The COM apartment fix is doing its job.

One cosmetic note: while idle the dropdown appends the probed size and shows `USB Video  640x480`, but the stream then negotiates 800x600. Both labels come from real data (`FormatLabel` appends the first advertised mode), it just reads as inconsistent. Not a finding, only mentioning it in case it is easy to align.

---

**Finding 3 - unplug detection and self-recovery: PASS**

Streamed for 18 min 12 s untouched with no false positive (15:24:23 to 15:42:35), then two unplug/replug cycles:

| | Cycle 1 | Cycle 2 |
|---|---|---|
| Detected at | 15:42:35.727 | 15:46:10.871 |
| Detector | FPS collapse - `measured 1.0 fps vs target 15` | Slow read - `Read took 1000 ms for 4 frames` |
| Time unplugged | ~2 min 00 s | ~1 min 20 s |
| Reopened at | 15:44:35.835 | 15:47:30.427 |
| Recovery after last failed retry | 4.6 s | 4.6 s |

Retries ran every ~2.25 s throughout, 86 attempts total, and never stopped. That is the behaviour that was broken before, so this is the important one.

Recovery was fully automatic both times - I did not touch the device dropdown.

**The longer unplug behaved exactly the same as the shorter one.** Same detection speed, same 4.6 s recovery, same steady retry cadence in between. That is the part I was most interested in, because the original wedge only showed up after a long absence - the retry loop stopping permanently. Nothing like that here at either duration.

To be straight about one thing: I intended a 3 minute absence on the second cycle but the log shows 1 min 20 s, so the longest gap actually tested was the 2 minutes in cycle 1. Both are well past the few seconds where the old build self-healed anyway, but I have not put a genuine 3+ minute gap through it. Happy to if you want it nailed down.

**Finding 5 - stale resolution in the banner: PASS**

While disconnected the banner shows `Disconnected` with no leftover `800x600 @ 14.8fps`.

Minor: the text shown is the reopen failure (`Could not open capture device index 0`) rather than the original stall reason, because the retry message overwrites it. The informative reason is still in the log, so this is only about what the user sees.

**Finding 6 - panel reconnects without a page reload: PASS**

Picture came back on its own after both replugs with no page reload and no manual intervention. `/api/video/status` confirmed `viewers=1` with `frameSeq` advancing, so the browser side MJPEG connection genuinely re-established rather than the banner alone updating.

**Finding 1 - Ctrl+F5 while streaming: PASS**

Host stayed up.

---

**Observation 1: the two detectors are complementary, and only one works headless**

The two cycles were caught by different detectors - cycle 1 by FPS collapse, cycle 2 by the slow-read watchdog. Both give ~4 s detection, so in practice it does not matter which fires.

But FPS collapse is inside the `ViewerCount > 0` check, and slow read is not. So with no browser attached, detection depends on the slow-read path happening to fire - and cycle 1 shows it does not always. On a headless or Pi setup, where nobody is watching most of the time, an unplug could sit undetected. Might be worth moving the FPS-collapse check outside the viewer gate, or lowering the empty-frame threshold when there are no viewers.

**Observation 2: the Media Foundation passthrough path cannot run**

MF enumeration fails on my machine and always falls through to OpenCV:

```
Radio Display: MFEnumDeviceSources returned hr=0xC00D36E6, count=0
```

`0xC00D36E6` is `MF_E_ATTRIBUTENOTFOUND` (Mferror.h). The `SetGUID` before it succeeds, so the attribute is being set under the wrong key. Four GUID constants in `Services/Video/WindowsMfMjpegSession.cs` do not match the Windows SDK:

| Constant | In code | Windows SDK |
|---|---|---|
| `MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE` | `c60ac5fe-672d-41ef-afc3-1f319d7f80b0` | `c60ac5fe-252a-478f-a0ef-bc8fa5f7cad3` |
| `..._SOURCE_TYPE_VIDCAP_GUID` | `8ac3587a-4ba1-4d9f-abb4-946d5be8add6` | `8ac3587a-4ae7-42d8-99e0-0a6013eef90f` |
| `MF_DEVSOURCE_ATTRIBUTE_FRIENDLY_NAME` | `60d0e559-52f8-4fa2-87e0-b834d243a97f` | `60d0e559-52f8-4fa2-bbce-acdb34a8ec01` |
| `MF_READWRITE_DISABLE_CONVERTERS` | `98d44c05-8a0d-4b80-8f1a-6f9a315fad27` | `98d5b065-1374-4847-8d5d-31520fee7156` |

`..._SYMBOLIC_LINK`, the four `MF_MT_*` keys and the media subtypes are all correct.

The first row is the fatal one - with the wrong `SOURCE_TYPE` key, enumeration fails on any machine, not just mine, so the MJPEG passthrough path has never actually run and its CPU saving is not being realised. My stream is doing the full YUY2 to JPEG re-encode (`uncompressed 14.4 MB/s vs ~22 MB/s USB2` in the log).

Worth noting the log's hint about camera privacy only applies when `hr` is 0, which it is not here - that sent me down the wrong path for a bit.

I have not changed anything, they are yours to fix. Say the word if you would rather I push the GUID correction to the branch so you can test whether passthrough works once enumeration succeeds.

mm5agm:
Correction to Observation 1 in my previous comment - please disregard it.

I said the `ViewerCount > 0` gate on the FPS-collapse detector could leave an unplug undetected on a headless or Pi setup. That is wrong, and I should have checked the viewer lifecycle before concluding it from the gate alone.

`IdleReleaseDelayMs` is 2000, and two seconds after the last viewer disconnects `ScheduleIdleRelease` calls `StopLoopAndWait()` - the capture loop exits and the device is released. Capture only starts again when `ViewerCount > 0`. So there is no state where capture is running with no viewers attached: with nobody watching, the device is closed and the status is `idle`, not a stale `streaming`.

That makes the gate harmless. It is only ever evaluated while a viewer is attached, which is the only time the loop runs at all. No change needed, and please ignore my suggestion about moving the check outside the gate or altering the empty-frame threshold.

The rest of the comment stands. In particular the two unplug cycles really were caught by different detectors - FPS collapse on the first, the slow-read watchdog on the second - both at about 4 seconds, which I still think is worth knowing.

Apologies for the noise.

fabiojvalente:
@mm5agm UI for video + image quality completed, please review :)

mm5agm:
Follow-up on the one caveat I left in my retest: I have now done a 5 minute unplug, which was the duration I said I had not properly tested.

Detected at 16:05:07.702 by FPS collapse, device back at 16:10:15.089 - 5 min 07 s out. 135 retries during the absence, every ~2.24 s, never stopping. Recovery 4.6 s after the last failed retry, identical to the 2 min and 1 min 20 s cycles. Ctrl+F5 afterwards was clean again.

Preceded by a 17 min 37 s clean stream with no false positive.

So absence duration makes no difference to recovery, including well past the point where the old build wedged permanently. Finding 3 is fully verified as far as I can take it.

mm5agm:
Reviewed `7691c11` on Windows 11, FTdx101MP front panel via MS2109 dongle. Builds clean (0 warnings, 0 errors). Thanks for doing USER_MANUAL 6.9 and 19 properly rather than leaving them to me - that is the part I usually end up rewriting.

**Finding 6 re-verified on this build: PASS**

The reconnect path picked up a new `wantStream` gate, so I re-ran one unplug/replug rather than assume my earlier results carried over:

- unplug detected 16:28:57.502 (`measured 1.0 fps vs target 60`)
- device reopened 16:31:06.818, about 2 min 09 s out
- picture returned on its own, no page reload, `viewers=1` with `frameSeq` advancing

The capture service itself is unchanged in this commit, so findings 3 and 5 stand from my earlier run.

**Image quality works - worth saying, because it does not look like it**

I could see no difference by eye at any preset and nearly reported it broken. Measuring MJPEG stream bitrate shows the encoder is honouring it:

| Preset | Stream bitrate |
|---|---|
| Low (50) | 1347 KB/s |
| Medium (65) | 1626 KB/s |
| Max (85) | 2411 KB/s |

A clean 1.8x spread. On an 800x600 radio panel the visual difference is just marginal unless you enlarge the image. No change needed - noting it so the next person who says "the quality selector does nothing" gets an answer.

**fps: the selector is inconsistent, and I cannot fully characterise it**

fps clearly does select the device mode at open time - that part is solid:

| Sequence | Negotiated | Measured |
|---|---|---|
| Reopen at target 60 | 640x480 | 59.5 |
| Reopen at target 15, after ~2 min idle | 800x600 | 14.8 |
| Reopen at target 15, immediately after | 800x600 | 19.8 |
| Live change 15 -> 60 while streaming | 800x600 | 19.9 |
| Live change 60 -> 15 while streaming | 800x600 | 19.9 |

Two things I cannot explain:

- At 800x600 the same target of 15 gives either 14.8 or 19.8 depending on what happened before. Not deterministic in my hands.
- Changing fps while streaming does not appear to take effect. Raising it does not renegotiate the device mode (expected), but lowering it again does not throttle the output back down either. A stop/start does apply the new value.

The practical consequence is that a Pi user who selects 15 to save CPU may still get ~20 until they restart the stream, which works against the guidance in section 19.4.

A guess, and only a guess, since you know this code far better than I do: the pacing guard at `VideoCaptureService.cs:670`

```csharp
if (readMs < frameInterval.TotalMilliseconds * 0.6)
```

only paces when Read returns early. Once the device is returning ~50 ms reads, that condition is false for every target, so the output just follows the device clock. That is the guard your earlier "30 fps locking at 19.8" commit added, so it may have traded the overshoot for the ability to throttle down. Please treat that as a hypothesis - I could not reproduce cleanly enough to stand behind it.

**"Click Start" names a control that is not visibly there**

The badge and placeholder say `Idle: click Start`, but `radioDisplayBadge` is a plain `<span class="badge">` and is not clickable - I clicked it first and nothing happened. The real control is `radioDisplayStreamBtn`, an icon-only green play triangle further right in the header, past the device dropdown, refresh button, and the fps and quality selectors. It has an `aria-label` and a tooltip, but no visible text.

I only found it by reading the markup. Several people who use this app are partially sighted or navigate with head-tracking and an on-screen keyboard, and an instruction naming a control with no visible matching label is a real obstacle. Could the button carry a visible `Start` / `Stop` label beside the icon? That would also make the instruction self-consistent.

**Minor**

- `macos_video_investigatin.md` is committed at the repo root, and the filename has a typo. Was that meant to ship? If it is worth keeping I would move it under `docs/`.
- `VideoJpegQualityOptions.Normalize` snaps to 50/65/85, where the old code clamped to 40-85. Anyone who had hand-tuned `VideoJpegQuality` below 50 on a Pi silently moves up to 65 on upgrade, which is the wrong direction for that case. Worth considering 40 for Low.
- The XML doc on `VideoJpegQuality` lost its allowed-range note and now just reads "Image Quality".
- At 640x480@60 the log reads `negotiated 640x480 }?6? device=60fps` - the driver is not returning a printable FourCC there, so `FormatFourCc` sanitises it to punctuation. Same line warns `uncompressed 36.9 MB/s vs ~22 MB/s USB2`, but the stream sustained 59.5 fps quite happily, so that warning looks like a false alarm. Cosmetic either way.

**One for the release notes rather than for you:** the stream no longer starting until Start is clicked is a sensible change, but existing users will open Home to a black panel after upgrading. I will make sure that is called out.

fabiojvalente:
@mm5agm please check `feature/add-video-tx-read-fps-res-from-source` branch

I've taken care of the minor items there (for RnD purposes and remove noise in here)

**JPEG quality** — Low is 40 again (old 40–85 floor). Hand-tuned values below 50 snap to nearest preset, not silently up to Medium 65. Restored the allowed-range note on VideoJpegQuality. (will follow up later on a separate PR on using a lossless format)

**Negotiate log** — unprintable FourCC logs as hex/unknown instead of }?6?. The USB2 “36.9 MB/s vs ~22 MB/s” line only warns for known uncompressed formats, so MJPEG 640×480@60 isn’t a false alarm.

**FPS selector** — we query the pin’s advertised max (V4L2 / AVFoundation / MF). Dropdown stays 15 / 30 / 60; anything above the device max is hidden. Still a target — USB/encode/CPU can deliver less. Saved 60 is kept in settings and should come back if you plug in a 60 fps stick later.

**click start** rephrased the text on the video window

Edit: 

I have pushed new commits into the `feature/add-video-tx-read-fps-res-from-source`

I noticed that even on mac, on some occasions the capture resolution set by the dongle wasn't always the same, in some cases it would do 600x480, in others 800x600 (windows here would stick to 600c480) this would led to the so known framerate instability we've been fighting

So I've check OBS (streaming tool) and was able to reach 60fps with 800x600 as expected, and as such decided to dig into the problem:

TL,DR: the poor image quality and the blurred digits you reported were indeed a symptom of the problem. I have now modified both macos and windows (on this separate branch) to prefer 800x600 @ 60 fps 

Radio Display now **reads FPS (and pin size) from the capture device** instead of offering a fixed list, then **opens the same panel-sized pin for 15 / 30 / 60**. On Windows it prefers the **compressed MJPEG pin** (the OBS path) so USB2 HDMI dongles can actually hit 30 fps. JPEG quality default is **Max (85)** so a dongle JPEG is passed through instead of recompressed.

## What landed

**Device-advertised rates → UI**

- `GET /api/video/devices` and `/status` now return `rates` / `maxFps`.
- The FPS dropdown is still **15 / 30 / 60**, but options above what the stick advertises are hidden (e.g. no 60 on a 30 fps dongle).
- Caps are queried per platform (V4L2, AVFoundation, Media Foundation) and cached in `VideoDeviceFpsCaps`. While capture is running, Windows does **not** probe-open the camera again (that was killing the host on stop / pop-out).


**One capture size for all FPS**

- Shared ranking in `VideoCapturePinRank` (mirrored in `MacAvFoundationFps.m`): prefer a **4:3 (or 5:3) pin ≥800 px wide**, typically 800×600 after scale. Changing FPS must not jump 640×480 ↔ 720p ↔ 1080p.
- **640×480** is last-resort only. If that 4:3 pin cannot do 60, size stays put and duration is clamped.

**Windows: DirectShow MJPEG passthrough**
- New `WindowsDshowMjpegSession` + `IJpegCaptureSession`: copy JPEG bytes off the MJPEG pin, no YUY2 decode.
- Capture order: DirectShow MJPEG → Media Foundation MJPEG → OpenCV fallback.
- That is the USB2 ~20 fps fix: uncompressed YUY2 at 800×600 does not fit the bus; OBS was already on MJPEG.

**macOS: native frame-duration control**
- `libYwcMacAvFps.dylib` (`MacAvFoundationFps.m`) sets `activeVideoMin/MaxFrameDuration` from the format’s real range (avoids `CMTimeMake(1, 60)` aborting on 59.94 HDMI pins).
- Async JPEG encode on the AppKit path; software pacing still applies when the pin floor is ~20 and the operator asked for 15.

**JPEG quality**
- Default **65 → 85 (Max)**. Max keeps the capture JPEG (least CPU). Low / Medium recompress.

**Docs / UI**
- `USER_MANUAL.md` §6.9 / §19 updated for advertised rates, shared pin, and Max-as-passthrough.
- Placeholder copy + icon; `video_debugging.md` is the investigation handoff (USB2 YUY2 vs MJPEG).

Please have a test, focused mainly on frame rates and image quality, if you're happy I'll rebase into this branch and look at the other points you mentioned earlier (in case they still apply)

mm5agm:
Tested `feature/add-video-tx-read-fps-res-from-source` this morning on an MS2109 dongle (VID_534D&PID_2109) fed from an FTdx101MP. Short version: the frame rate work is a clear win and the image quality is good. I found one bug I think should be fixed before this rebases, plus a few smaller things.

## What works

Both TFMs build clean here (0 warnings, 0 errors), which the previous branch did not.

All three presets negotiate the same capture size and stay there, with no size jumping between rates:

| Target | Measured | KB/frame | Wire rate | CPU (1 core) |
|---|---|---|---|---|
| 15 | 15.0 | 88 | 1.26 MB/s | - |
| 30 | 29.5 | 86.3 | 2.49 MB/s | - |
| 60, quality Max (85) | 56.2 | 86.2 | 4.73 MB/s | 13% |
| 60, quality Medium (65) | 56.4 | 67.2 | 3.70 MB/s | 23% |
| 60, quality Low (40) | 55.8 | 51.4 | 2.80 MB/s | 21% |

15 fps pacing is now exact (60 frames in 4.0 s), so the old "15 shows about 20" is gone.

The dongle's pin list shows why the shared pin ranking matters here:

```
MJPG 1920x1080@60 ... MJPG 800x600@60 ... MJPG 640x480@60
YUY2 800x600@20, YUY2 720x480@30, YUY2 640x480@30
```

MJPG offers 60 at every size, YUY2 800x600 caps at 20. Picking the MJPG pin is what removes the 20 fps ceiling.

Image quality: the FTdx101MP frequency digits were crisp and sharp throughout. That was the original complaint and it is fixed.

Two things worth knowing about the quality settings, from the numbers above. Max as the new default looks right: Medium buys only 22% less data for roughly double the CPU. But Low and Medium cost about the same CPU (21% vs 23%, which is inside the noise here) while Low sends 24% less than Medium and 40% less than Max. The expensive part is clearly the decode, not the re-encode, so once a re-encode is happening at all there is little reason to stop at Medium. That makes Low the useful setting for anyone running the panel over a VPN, and the frame rate held at about 56 fps at every quality, so none of this costs frames.

## Bug: fallback paths latch off permanently after an unplug

This is the one I would like fixed before rebase.

`_dshowMjpegUnavailable` and `_mfUnavailable` in `VideoCaptureService` are set on any failed open, including "the device is not there any more". They are cleared only in the Stop path. So an unplug/replug cycle silently disables both fast paths for the rest of the capture session.

Reproduced end to end:

```
08:12:49  800x600 MJPG  mode=passthrough, via=DirectShow    56 fps
          (dongle unplugged; retry loop now sees a different camera at index 0
           with no MJPEG pin, which sets both latches)
08:19:50  streaming device index 0 ... passthrough=false
08:19:51  800x600 YUY2  mode=encode                         20.2 fps
```

So after a replug you are back on the exact 20 fps path this branch exists to remove, with no indication in the UI. Pressing Stop then Start restores it immediately on the same device and same settings:

```
08:23:29  Radio Display capture stopped
08:23:31  800x600 MJPG  mode=passthrough, via=DirectShow    56 fps
```

I understand the intent of the comment at the `_mfUnavailable` assignment ("do not re-probe on every reconnect once we know this host's device has no ranked MJPEG pin") - the probe is expensive. The problem is that the latch cannot distinguish "this device has no MJPEG pin" from "no device". Suggested shape: only latch when the device opened successfully but exposed no usable MJPEG pin, never when the open itself failed, and reset both whenever the resolved device index changes.

## Unplug detection takes about 30 seconds, and logs nothing

Pulling the cable took roughly 30 seconds to show Disconnected on screen. The passthrough loop counts iterations (`EmptyFrameGiveUp` = 45) and each `TryReadJpeg` blocks for `ReadWaitMs` = 1000 ms, so the give-up threshold is a time in disguise. An elapsed-time check would be both faster and easier to reason about.

Related: the empty-frame `MarkDisconnected` has no log line, so the log jumps straight from healthy frames to "failed to open device index 0" with nothing in between. That made this harder to diagnose than it needed to be. Worth a warning there.

## Device rate capping never activates

`/api/video/status` reports `"maxFps":0,"rates":[]` at all times here, so the new device-advertised-rate capping never takes effect.

The ordering looks like the cause: `VideoDeviceFpsCaps.RatesFor(key, allowDeviceOpen: !_capture.IsCapturing)`. Before a device is chosen the key is empty, so nothing is queried; the moment one is chosen capture starts, and `allowDeviceOpen` is false from then on. Windows enumeration only peeks the cache and never populates it.

Invisible on this dongle because it advertises 60 everywhere, but a 30 fps device would still be offered a 60 preset.

## Accessibility: the placeholder icon has no accessible name

In `Index.cshtml` and `RadioDisplay.cshtml` the placeholder became:

```html
Click <i class="bi bi-play-fill" aria-hidden="true"></i> to show the radio display
```

With the icon hidden and no text alternative, a screen reader now reads "Click to show the radio display" and the control is never named. The previous wording named it. Several YWC users are partially sighted or use assistive input, so I would rather keep a name in there - a visually-hidden "Start" next to the icon would do it.

Same thing in `radio-display-ui.js`, where `panel.setStatus('idle', 'click Start')` was reduced to `panel.setStatus('idle')`.

## Minor

Device keys are bare indexes on Windows (`index:0`), so when the dongle disappeared another camera took index 0 and YWC spent two minutes trying to open the wrong device. The dropdown note already warns about this, so I mention it only as something to keep in mind rather than something to fix here.

While that was happening the retry loop logged a warning every 2.25 s indefinitely. Some backoff would keep the log readable.

Thanks for the docs updates in USER_MANUAL 6.9 and 19 - good to have those written alongside the code rather than left to me.

fabiojvalente:
Hi @mm5agm 

Thanks for the time taken testing this branch

on the Bug: fallback paths latch off permanently after an unplug:

Agreed this is a real failure, but the latch is a symptom of the retry loop treating index:0 as a stable device.

After an unplug the saved key is still index:N. Windows reshuffles the list, so the next open is often a different camera at the same index. That camera binds successfully, has no MJPEG pin, and both latches stick. When the dongle returns to the same index we stay on the YUY2 encode path with no UI signal. Stop/Start works because that is the only place the latches clear.

The suggested shape (latch only when the device opened with no MJPEG pin; reset when the index changes) does not cover this run: the other camera did open with no MJPEG pin, and the index never changed.

**I don’t think its reasonable to chase silent unplug/replug recovery**. USB video indices are not identity, a session-wide boolean cannot mean “this physical dongle has no MJPEG pin,” and operators are not hotplugging the capture dongle. It gets plugged in at the station and left there. A cable knock or USB glitch should fail clearly, not try to be clever and land on the laptop camera at 20 fps.

Intended fix: on unplug, stay **Disconnected** and do not open whatever is now at that index. Replug, refresh the device list, Start. That is already the reliable recovery, and it avoids streaming the laptop camera in the gap. I’ll drop the “reattaches without a reload” claim in the manual to match, and suggest the user refresh the device list before restarting the stream.

Happy to distinguish “bind failed / device gone” from “no MJPEG pin” while doing that, but I would stop trying to auto-adopt a stranger camera/capture device just to preserve the fast-path latch.

Let me know your thoughts.

I'll chase the other ones in the meantime and report back.

mm5agm:
Hi @fabiojvalente

You are right and my suggested fix was wrong. Both of my conditions fail on the run I
actually logged: the other camera opened successfully, it had no MJPEG pin, and the
index never changed. Neither "only latch on a successful open" nor "reset when the index
changes" would have caught it. Your read of the root cause is the correct one - the latch
is a symptom, and the real fault is treating `index:N` as identity.

## Staying Disconnected

Agreed, and I think it fixes a worse bug than the one I reported. Twenty fps is a
degraded picture. Silently streaming whatever camera happens to be at index 0 into a
panel labelled "radio display" is a wrong picture, and on a laptop that is somebody's
webcam. Failing clearly is the better behaviour and I am happy for the branch to do that.

One thing to flag so it is a deliberate trade rather than a silent regression: automatic
recovery is not just a manual claim, it is a finding I tested and passed on 14 August -
four unplug/replug cycles, recovery about 4.6 s after the last failed retry, no page
reload, no touching the dropdown. It passed because the dongle came back at the same
index with nothing competing for it. Take the behaviour out by all means, but it is
tested behaviour going out, so dropping the manual line is exactly right, and the PR
description promises it too.

## The identity option, as a follow-up and not a blocker

The reason auto-recovery had to be either clever or absent is that the saved key carries
no identity. Both Windows capture APIs will give you one: DirectShow has the device path
(`\\?\usb#vid_534d&pid_2109#...`) and Media Foundation has the symbolic link - and
`MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_SYMBOLIC_LINK` is already sitting in
`WindowsMfMjpegSession.cs` as a constant, with the correct value.

Key the saved device on that instead of an index and three things follow at once: the
retry loop cannot bind a stranger because the identity check fails; a replug that lands
at a different index is still found; and auto-recovery becomes safe rather than clever,
because "is this the same physical device" becomes a question you can actually answer.

I am not asking for it in this PR. It changes the `VideoCaptureDeviceKey` format and
needs a migration for anyone who has already saved one, which is scope growth on a branch
that is already large. Your policy fix is the right thing to ship now. I mention it
because it is the change that would let the recovery come back later without any of the
guessing, if either of us wants it.

## Three things that stand whichever way the policy goes

You said you would chase the others, so just to be clear which ones survive this
decision, since they are all about what the operator and the log can see:

1. Once we have decided to stay Disconnected, the retry loop should stop or back off. At
   the moment it warns every 2.25 s indefinitely, and with the new policy those retries
   are not even trying to achieve anything.
2. `MarkDisconnected` on the empty-frame path still logs nothing. The log jumping from
   healthy frames straight to "failed to open device index 0" is what made this one hard
   to pin down.
3. The give-up threshold is still `EmptyFrameGiveUp` = 45 iterations x `ReadWaitMs` =
   1000 ms, which is a 30 s timeout written as a counter. An elapsed-time check would be
   quicker to reach and easier to read.

## Separate: four wrong GUID constants in WindowsMfMjpegSession.cs

Unrelated to the above, and I should have raised it before now. Four of the GUID
constants do not match the Windows SDK. I have just checked them against
`mfidl.h` / `mfreadwrite.h` from SDK 10.0.26100.0 on this machine:

| Constant | In the branch | Windows SDK |
|---|---|---|
| `MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE` | `c60ac5fe-672d-41ef-afc3-1f319d7f80b0` | `c60ac5fe-252a-478f-a0ef-bc8fa5f7cad3` |
| `MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_GUID` | `8ac3587a-4ba1-4d9f-abb4-946d5be8add6` | `8ac3587a-4ae7-42d8-99e0-0a6013eef90f` |
| `MF_DEVSOURCE_ATTRIBUTE_FRIENDLY_NAME` | `60d0e559-52f8-4fa2-87e0-b834d243a97f` | `60d0e559-52f8-4fa2-bbce-acdb34a8ec01` |
| `MF_READWRITE_DISABLE_CONVERTERS` | `98d44c05-8a0d-4b80-8f1a-6f9a315fad27` | `98d5b065-1374-4847-8d5d-31520fee7156` |

The first one is fatal on any machine: `SetGUID(SOURCE_TYPE)` writes an attribute nobody
reads, so `MFEnumDeviceSources` returns `0xC00D36E6` (`MF_E_ATTRIBUTENOTFOUND`) and the MF
path falls through to OpenCV every time. The MF_MT_* constants and the symbolic-link one
are all correct, so it is these four only. In the first three the opening 32 bits are
right and the rest is not, which looks like they were reconstructed rather than copied.

Priority on this is much lower than it was, and I want to be straight about why: your
DirectShow work has made MF a fallback that never fires on Windows. My run on 15 August
was `mode=passthrough, via=DirectShow` at 56 fps and 13% CPU, so the passthrough win is
real and it is being delivered by the DirectShow path. Fixing the GUIDs would only matter
on a host where DirectShow cannot rank an MJPEG pin. Worth correcting whenever you are
next in that file rather than something to hold anything up for. Happy to push the four
values if you would rather not touch it.

73 Colin MM5AGM

fabiojvalente:
Hi @mm5agm

Fixes have been pushed and its ready for one more batch of tests :)

On unplug (or a failed open) Radio Display goes Disconnected and stays there. It does not reopen `index:N`, so it will not pick up whatever camera Windows has put in that slot. 
**Recovery is**: 
1. refresh the device list
2. Start. 

Auto still means start when the panel opens, **not retry after an unplug.** The same policy is on macOS and Linux; the host loop is shared. I have updated USER_MANUAL 6.9 / 19.3 / 19.5 to match, including dropping the old “reattaches without a reload” claim.

Regarding the three items that stood either way:

- **The retry loop stops**. No more open-every-~2.25 s warnings. A fresh Start (or a device change in the dropdown) is the only reopen.
- **The empty-frame MarkDisconnected path now logs a warning with the reason**, same as the stall / FPS-collapse paths.
- **Give-up is wall-clock, ~5 s of no frames, not 45 × 1 s reads.**


73 Fabio

mm5agm:
Hi @fabiojvalente

Tested 3318ecd on the FTdx101MP front panel through the MS2109 dongle. The three items
you listed all pass, and the latch bug I originally reported is fixed. Three new things
below, none of them showstoppers.

A soak test is running now. I will report back on that in about an hour.

## What passed

| Check | Result |
|---|---|
| Unplug to Disconnected | 5111 ms, was 30 s |
| Give-up logs a reason | Yes, with the elapsed time |
| Retry loop stops | Silent for the full 1m50s I left it unplugged |
| No auto-recovery on replug | Correct, nothing happened until I pressed Start |
| Refresh list + Start recovers | Yes |
| MJPEG fast path restored | Yes, `mode=passthrough, via=DirectShow` |

The last row got a harder test than I planned, and it is worth spelling out because it is
exactly the failure I reported. Partway through, a stranger camera really did get selected at
`index:0` - OBS Virtual Camera, which reports no MJPEG pin, so it set
`_dshowMjpegUnavailable` for real. The next capture start cleared the latch, re-probed, and
came back on the DirectShow passthrough path rather than the 20 fps encode path:

```
11:56:49.277 [WRN] DirectShow no JPEG for 5111 ms - treating as unplug
11:56:53.767 [INF] capture stopped
             (silence - no retry warnings)
11:58:43.471 [INF] 'OBS Virtual Camera' DirectShow has no MJPEG pin - falling back
11:58:43.776 [WRN] failed to open device index 0
11:59:38.346 [INF] negotiated 800x600 MJPG device=60fps (mode=passthrough, via=DirectShow)
```

So the latch fix is not just theoretically right, it survived the real thing. Thank you -
your read of the root cause was better than mine and the fix is cleaner than what I asked
for.

## 1. The halt is browser state only, so a viewer attaching walks past it

Not a blocker, and the new policy is a clear improvement on what was there before. But it
is worth knowing the guarantee has an edge, so we can decide whether to close it now or
note it and move on.

`holdDisconnected` lives in `radio-display-ui.js` as a module variable, and server side
`RunCaptureLoop` clears `_haltAfterDisconnect` at the top of every start. That is
reasonable in itself - the service cannot tell "operator pressed Start" from "a viewer
attached". The consequence is that the policy only holds for as long as that one page
stays loaded.

I reproduced this with no browser involved at all. With the dongle unplugged and the panel
sitting in its held Disconnected state, a bare HTTP GET to `/api/video/stream` was enough
to make the host reopen `index:0` and bind OBS Virtual Camera. That is the 11:58:43 line
above. It only stayed safe because OBS was not running, so the open failed. Had it been a
real webcam it would have opened and streamed somebody's face into a panel labelled Radio
Display, which is the outcome the new policy exists to prevent.

The everyday route to the same place, and this part I have read in the code rather than
measured, is a page reload with **Auto** ticked: `radio-display-ui.js:692` restores
`wantStream` from `localStorage` and starts the stream, while the fresh page has
`holdDisconnected = false`. So "refresh the page" would appear to users as a working
recovery that quietly rebinds whatever is at that index. I have not tested that path
directly, so treat it as a code reading rather than a measurement.

If you want to close it, something server side would need to remember the halt across a
viewer attach. If you would rather leave it, that is a fair call too - it is a narrower
hole than the one this commit just filled, and it needs a reload or a direct API hit to
reach. I would just want us to have chosen it rather than inherited it.

## 2. The device dropdown keeps the dead device's name

While disconnected, the dropdown still read "USB Video" - the dongle that was no longer
plugged in. My first instinct was that it should read "(None)".

`pollStatus` only reloads the device list when the select value stops matching the saved
key (`radio-display-ui.js:468`). On unplug the select still holds `index:0` and so does
`currentDeviceKey`, they match, and it never re-fetches. `onDeviceLost` does not refresh it
either.

But calling `loadDeviceSelect` there would not give "(None)". The orphan branch at
`radio-display-ui.js:355-362` that renders "(not present)" only fires when the saved key is
absent from the list, and `deviceOptionMatches` compares on index alone. With the dongle
out, `/api/video/devices` returned:

```json
{"key":"index:0","label":"OBS Virtual Camera","index":0}
```

`index:0` is present, so the refresh would have relabelled the dropdown "OBS Virtual
Camera" and left it selected as the radio display device. That is arguably worse than the
stale label. On any machine with a second camera - which is the machine where this bug
matters at all - the "(not present)" branch can essentially never fire.

I am not asking for the identity change in this PR, I still think that is a follow-up. I am
raising it because it is the same root cause showing up somewhere a user can see it, rather
than an architectural argument. A short-term option is to grey the dropdown out and show
the held error while `holdDisconnected` is set, so at least it stops asserting something
untrue.

Small related thing: I recovered by reselecting "USB Video" and then pressing Start, but
reselecting an option that is already selected fires no `change` event, so `setDeviceKey`
never ran. Start alone did the work, because `requestStart` clears the hold itself. Worth
knowing since USER_MANUAL 19.3 now tells users to refresh the list and then Start - in the
same-index case the refresh step does nothing.

## 3. The halt now applies to the two stall heuristics as well

`MarkDisconnected` sets the halt flag, and it is called from more than the unplug path:

- `VideoCaptureService.cs:518` - fps collapse, measured below 25% of target for 3 windows
- `VideoCaptureService.cs:679` - slow read, `readMs >= 750` on 4 consecutive frames

Both of those were written as reopen triggers. The comment immediately above the slow-read
one still says so:

```csharp
// Unplug: DirectShow keeps succeeding with stale frames
// but Read blocks (~1 fps). Break out so we reopen.
```

That comment is now false - it breaks out and stays broken out. So a transient USB hiccup
or a momentary fps dip that used to self-heal now takes the panel down until the operator
presses Start.

I do not know how often those heuristics trip in normal running, which is why I am soaking
it rather than asserting there is a problem. It may be that they only ever fire on genuine
unplugs, in which case this is just a stale comment to fix. But if they can fire
spuriously, "stays Disconnected" turns a self-healing glitch into an operator callout, and
that is a different trade from the one we agreed. Worth deciding deliberately either way.

## The GUIDs, briefly

Not asking you to act, just noting the evidence turned up on its own. This fired on my
machine during the run:

```
MFEnumDeviceSources returned hr=0xC00D36E6, count=0
(MF_E_ATTRIBUTENOTFOUND - SOURCE_TYPE did not stick; will try DirectShow DevicePath).
```

That is your own message from cb5e5dc, two days before I raised the constants, so you had
already found the same fault and routed around it with the DevicePath fallback. Fixing
`MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE` would let the enumeration work rather than needing the
fallback, but with DirectShow passthrough carrying Windows it stays low priority. Offer to
push the four values still stands if you would rather not touch that file.

## Not part of this commit, still open

`maxFps` and `rates` come back empty in `/api/video/status` throughout, so the device rate
capping never activates and the fps dropdown never learns what the dongle supports. I think
that is the `RatesFor(..., allowDeviceOpen: !IsCapturing)` ordering. Separate from this
change, just confirming it is still there.

The fps selector inconsistency I reported earlier has not reappeared on this run - 14.8
measured against a target of 15, sustained over four minutes, which is correct. I am not
calling it fixed, only saying it did not show today.

73 Colin MM5AGM

mm5agm:
Separate from the test results, and this is good news rather than a finding.

While the soak has been running I have been driving the radio's scope from the app - span,
ref level, the MAIN/SUB buttons - using the CAT scope control that went in on #109. Those
changes come back to me on your video stream, and it works very well.

Put together, the two features are a remote panadapter with no SDR in the path at all. I
change the span in the browser and watch the radio's own scope respond in the browser. That
is a capability I had been chasing down a much harder road: trying to decode the FT-710's
SCU-LAN10 spectrum stream, which is undocumented, has no known wire format and no tester.
This gets to the same place with a cheap HDMI dongle and CAT commands that are already
bench-verified, and it works on any radio with a scope on its front panel rather than
needing per-model reverse engineering.

I mention it because I think your PR description undersells the feature. "See your radio's
display remotely" reads as a convenience. Paired with scope control it is closer to a
capability - the operator gets spectrum on a radio YWC has no SDR path for, and on a Pi
where an SDR would be a real cost. Worth a line in the manual too, which I will write.

Nothing for you to do here. I thought you would want to know the feature turned out to be
worth more than the sum of its parts.

73 Colin MM5AGM

mm5agm:
One more thing your video stream turns out to be good for, which I had not expected.

It is not only the scope. Changing the antenna, the attenuator, the AMP, the roofing filter
or the AGC from the app shows up on the radio's own display immediately, on your stream, as
the command lands.

That is worth more than it sounds. For most of those controls YWC confirms a change by
reading the setting back from the radio afterwards - which is really the radio agreeing
with itself. If a command were silently ignored, or landed on the wrong receiver, or the
radio quietly clamped the value, the read-back would not necessarily tell me. The front
panel is outside that loop, so your stream is the first genuinely independent confirmation
the app has ever had that a control did what it said.

The practical use is bug reports. When someone tells me a control does nothing on their
model, the most useful thing they can send me has always been hard to get - proof of what
the radio itself did, not what the app believed. A few seconds of your stream is exactly
that, and it needs no test equipment and no CAT log.

I have written this into the manual alongside the panadapter note, so it goes out with the
feature rather than being folklore.

Nothing for you to do. I keep finding that this PR is worth more than its description says.

73 Colin MM5AGM

mm5agm:
Soak test finished. Clean result.

  Continuous run   11:59:38 to 13:10:58, 71 minutes
  Frames           about 60,300
  Rate             14.7 fps against a target of 15, no drift over the hour
  Mode             passthrough, via DirectShow, 800x600 MJPG
  Warnings/errors  none
  Reopens/stops    none

Three log lines from the capture service in the whole window, the last of them being the
successful negotiation. Nothing else to report, which is the right outcome for a soak.

The soak also settles my third point, though not the way I expected. Neither stall
heuristic fired at any time - no slow-read give-up, no fps collapse. So I cannot tell you
from this run whether halting on those two paths behaves correctly, because they never
ran.

But it answers the question that mattered more. Those heuristics were the real risk in
this commit: if either threshold were tuned even slightly too tight, it would declare a
false unplug during entirely normal running, and with the new halt in place that now stops
the stream dead until the operator presses Start. Before, a false positive cost a reopen
and the viewer probably never noticed. So the cost of getting those thresholds wrong went
up with this change. Seventy-one minutes of ordinary operation produced not one false
positive, which is the evidence I wanted. My third point therefore drops back to what it
honestly is - a comment that now contradicts the code it sits above - rather than a defect.

On behaviour, that is everything I can usefully test from here, and it all passes. The two
points still worth your time are the first two: the halt living only in browser state, and
the device dropdown keeping the dead device's name.

One last thing, unprompted. The picture quality is excellent - sharp, and to my eye exactly
what is on the radio's touchscreen. My SDRs still give me a better spectrum, which I think
is simply the radio drawing its trace into a few hundred pixels of a small panel, and not
anything to do with your capture path. It does not matter, because I am going to run your
panel permanently alongside them anyway. Not for the scope - for everything else on that
screen. It is the only place in the app where I can see at a glance what the radio itself
thinks its controls are set to.

Thanks for the work on this one.

73 Colin MM5AGM

fabiojvalente:
Good morning @mm5agm 
Give it another go, went through the 3 points you raised earlier

mm5agm:
Hi @fabiojvalente

Tested fede2cf. Point 1 is properly closed, and I could measure it this time rather than
argue it from the code. Point 3 I agree is just the comments. Point 2 is partly closed, and
I think the remaining half is the same root cause as before rather than anything new.

## The part I keep using

Before any of the detail. The thing I like most in daily use is being able to detach the
display into its own browser page. That turns out to be how I actually want to run it - the
radio's screen off on its own window or a second monitor, the control panel free of it, and
Reattach when I want it back in one place. It is a small thing in the diff and a big one at
the desk. Thank you for building it that way rather than nailing the video into the Index
panel.

The rest of the happy path is unchanged and good as well: 800x600, mode=passthrough,
via=DirectShow, halt cleared by a device change, stream clean.

One finding further down is a correction to something I told you, and it makes the GUIDs
worth more than I said they were.

## Method, in case it is useful to you

I did not need the dongle for the halt tests. A failed open calls MarkDisconnected the same
way an unplug does (VideoCaptureService.cs:471), so pointing VideoCaptureDeviceKey at a
non-existent index:9 engages the halt deterministically, on any machine, with no hardware.
That gave me a repeatable rig for the exact thing I could only produce by accident last time.

## 1. The halt now holds. Measured, both sides.

Server side, with the halt engaged:

| Check | Result |
|---|---|
| Bare GET /api/video/stream, no browser | HTTP 409, device not reopened |
| Repeated bare GETs (3 more) | 409 each time, still no reopen |
| POST /api/video/start | 200, halted=false |
| Stream GET after Start | Reopens, fails again, re-halts. Correct. |

That was my main worry and it is gone. The 409 body carries the error plus the recovery
instruction, which is a nice touch.

The route I had only read in the code, and never measured, was Auto plus a page reload. I
have now measured it. Chrome, ywc.radioDisplayAutoStart set to 1, page loaded while the
server was halted:

    /api/video/stream requests issued by the page: 0
    badge: "Disconnected: Could not open capture device index 9."
    fps / quality / Auto controls: disabled

Zero. The reload does not walk past the halt. Probing status before deciding wantStream in
initRadioDisplayUi is what does it, and it is the right place for it.

## 2. The dropdown - half closed, and the other half is the identity thing again

The genuinely-absent case now works. With index:9 saved and not present, the dropdown reads
"index:9 (unavailable)", which is honest.

The case I actually reported was a reoccupied index, and I do not think that one changed.
In deviceOptionMatches the exact key match returns before the new holdDisconnected guard:

```js
const key = d.key || '';
if (key && key === want) return true;   // returns here
if (holdDisconnected) return false;     // never reached in the reoccupied case
```

selectMatchesSavedKey has the same shape. So with index:0 saved and a stranger camera now
at index:0, the saved key is an exact match for the stranger, the guard never fires, and the
dropdown labels it with the stranger's name.

I could not stage that without unplugging my dongle mid-session, so please treat it as a
code reading rather than a measurement - I am not repeating last time's mistake of asserting
behaviour I have not seen.

I am still not asking for the identity change in this PR. What you did add is the mitigation
I suggested, and it works: fps, quality and Auto are disabled while held, and the device
select carries a tooltip explaining the recovery. Leaving the select itself enabled is right,
since choosing a different device is one of the two ways out.

## 3. The stall heuristics - agreed, it was the comments

You fixed the comments and left the behaviour. That matches where the 71 minute soak left it,
so nothing more from me.

## Correction: the four GUIDs are not cosmetic. They cause the empty rates.

I told you the MF constants were low priority because DirectShow carries Windows and the MF
path never fires. The first half is true. The second half was wrong, and I am the one who
narrowed it too far.

maxFps:0 and rates:[] - the finding I have been carrying as "still open, probably the
allowDeviceOpen ordering" - is not the ordering. On Windows the chain is:

    VideoDeviceFpsCaps.RatesFor
      -> QueryWindowsRates
        -> WindowsMfMjpegSession.QueryFrameRates
          -> Media Foundation

Both MF entry points on that path set MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE, and both use the
wrong value:

    WindowsMfMjpegSession.cs:304   EnumVideoDevicesQuiet
    WindowsMfMjpegSession.cs:561   CreateDeviceSource   (the DevicePath fallback)

So the enumeration fails, the fallback fails too, and QueryFrameRates returns an empty array
for every device, always. Confirmed in this run's log:

```
MFEnumDeviceSources returned hr=0xC00D36E6, count=0
(MF_E_ATTRIBUTENOTFOUND - SOURCE_TYPE did not stick; will try DirectShow DevicePath).
```

Result: /api/video/status and /api/video/devices report maxFps 0 and rates [] for every
device they list, the fps dropdown never learns what the stick supports, and the capping
never activates. It is a visible bug with a wrong constant behind it, not a dormant fallback.

Two ways to fix it, and DirectShow makes the second one tempting:

1. Push the four SDK values. Offer still stands if you would rather not touch that file.
2. Source the rates from DirectShow instead, which already enumerates them correctly. Your
   own negotiation log prints the full pin list for my MS2109, rates included:

```
'USB Video' DirectShow pins: MJPG 1920x1080@60, MJPG 1600x1200@60, ... MJPG 800x600@60,
... YUY2 800x600@20, YUY2 640x480@30 ...
```

Everything the dropdown needs is in that list, on the path that actually works on Windows.
Your call which route - I do not mind doing (1) either way.

## The tests

Really glad to see these. Pulling the halt out into its own class and injecting it through
CreateService is what made the service testable at all, and AcquireViewerAsync_RefusesWhenHalted
is a proper test of the policy - it would fail if the guard came out.

One note for whenever you next touch that file, not now. The three in VideoDisconnectHaltTests
that describe the policy in their names - ViewerAttach_DoesNotClearWithoutOperatorAction,
OperatorStart_ClearsHalt, DeviceChange_ClearsHalt - all exercise the flag directly rather than
the paths they are named after, so they would still pass if /api/video/start or
/api/video/device stopped clearing the halt. The good news is that the behaviour is reachable
now that the seam exists: AcquireViewerAsync throws while halted, EnsureLoopStarted returns
early, and both controller actions call ClearDisconnectHalt. Cheap to point them at that next
time round.

Also worth knowing, since this is the first C# test project in the app outside core/, it sets
the convention. core/tests/RadioWebControl.Core.Tests is the other one if you want them to
match.

## Build

Both target frameworks clean on Windows, 0 warnings 0 errors, net10.0-windows and net10.0.
The Tests/** exclusions in the csproj are needed and correct - Tests/ already held a
PowerShell suite, so that folder was previously never compiled.

## Two things to decide, neither blocking

**A failed first open now latches too, not just an unplug.** MarkDisconnected is on the
open-failure path, so an open that fails for a transient reason latches the halt exactly like
an unplug does. On a Pi that autostarts, if YWC comes up before the USB capture device has
enumerated, the panel is dead until somebody presses Start - and on a headless remote station
that is a callout. I think that is a consequence of the policy rather than the intent of it.
An open that has never succeeded is a different event from a device that vanished mid-stream,
and it might be worth letting the first one retry a few times before it latches.

**fps, for the record and not as a claim.** I measured 12.7 against a target of 15 in a real
browser this time, over a 20 second frameSeq delta, badge agreeing. The soak on 3318ecd gave
14.7 at the same settings on the same dongle. I am not calling that a regression - fede2cf
touches no pacing code and my host was busy running a build and a headless Chrome - but it is
a third value for the fps inconsistency that is still open, after 14.8 and 19.8.

73 Colin MM5AGM

fabiojvalente:
Hi @mm5agm

Happy to leave identity, first-open latch, and the fps scatter as follow-ups — none of those belong in this PR any more.

The one leftover that is still a visible bug is empty `maxFps` / `rates` on Windows (the MF GUIDs). Do you want that fixed here before merge, or shall we merge as-is and I open GitHub issues for the pending items so they do not get forgotten?

If we go the issue route I would file at least:

- empty FPS rates / MF GUID constants (or DirectShow pin list for the dropdown)
- stable device identity (reoccupied index still labels the stranger camera)
- first-open failure latching the halt (Pi / autostart before USB is ready)

I also have the scope-controller work mostly ready for discussion/review, and it is waiting on this one landing.

73 Fabio

mm5agm:
Hi @fabiojvalente

Merge as-is and file the issues. This PR set out to fix the halt and the
recovery, it does that and I have measured it doing it, so it should land on
that basis rather than growing a fourth topic on the way in.

Two amendments to the plan though.

**The MF GUIDs should not go on the backlog.** File the issue by all means so it
is tracked, but I would rather take your offer and have it as its own small PR
straight after this one merges, not as a queued item. It is four constants, you
have offered twice now, and unlike the other two it is visible to every Windows
user today - the fps dropdown never populates for anyone, on any device. That is
the sort of known fix with a willing author that quietly rots if it goes in a
list.

**Route 1, the SDK values.** Sourcing the rates from the DirectShow pin list is
probably the better end state, since DirectShow is what actually carries Windows
and it already enumerates them correctly. But it moves where the data comes from,
and it leaves the MF path broken for whoever needs it later. Fix the constants
first while the diff is four lines and easy to check. If MF turns out to stay
unused we can revisit sourcing from DirectShow later, on its own merits rather
than as a workaround.

Your three issues are the right three. Identity and the first-open latch are both
genuinely follow-ups, and I am content they are recorded rather than rushed - the
first-open one especially, because the right answer there is probably a retry
policy and that deserves thinking about rather than a quick guard.

## One thing to check before you go further on the scope controller

You mentioned scope-controller work waiting on this landing, and I want to flag
something before you spend more time on it rather than after.

I merged **#109, scope control over CAT**, on the 15th. It drives the radio's own
scope over the SS command - Center / Cursor / Fix placement, span, W/F and 3DSS,
size, Hold, Marker, reference level - and it follows changes made at the front
panel back into the browser. It is currently gated to the FTdx101, because that is
the only rig I could bench-verify SS on.

If your scope controller is the same ground, I would much rather you found that
out from me now than at review. If it is something adjacent - the SDR spectrum
side, or a refactor, or extending the gating to other models with a rig I do not
have - then it may slot straight in on top and I would like to see it. What does
yours cover?

Not a criticism of anything you have done. It is my fault for merging it while
your attention was on the video work, and it is exactly the kind of collision two
people working in parallel are going to hit now and then.

73 Colin MM5AGM

fabiojvalente:
@mm5agm I am merging this one and tracking the MF GUIDs  on a separate PR / Bugfix 👍

PR #99, opened 2026-08-13
Shared core, phase 1: subtree the core repo in and move DxSpot into it

mm5agm:
Phase 1 of the shared-core plan Colin and Fabio agreed by email on 2026-08-13. The plan itself lives in the IWC repo at [`docs/design/shared-core-plan.md`](https://github.com/mm5agm/Icom_Web_Control/blob/develop/docs/design/shared-core-plan.md); the short version is that IWC was cloned from YWC, so most of the code above the radio layer is the same code maintained twice — measured this month at **62 of 89 shared-path files effectively identical**.

New repository: **[mm5agm/Radio_Web_Control_Core](https://github.com/mm5agm/Radio_Web_Control_Core)**, public, GPL-3.0, pulled in here at `core/` as a **git subtree**.

## What actually moved

`Models/DxSpot.cs`, and nothing else. It was byte-identical to IWC's copy bar the namespace, which is exactly why it was picked — **the point of phase 1 is to find out whether the plumbing survives, not to share anything valuable.**

`Services/DxClusterService.cs` is the only C# file in this repo that used the type. Everything else matching "DxSpot" is the unrelated `DxSpotAgeMinutes` setting or browser code.

## The bit that matters for this repo specifically

The core targets **`net10.0`, not `net10.0-windows`**, and the `ProjectReference` is deliberately **not** conditioned on `TargetFramework`. This repo multi-targets for the macOS/Linux CAT-only host, and a Windows-only core would have built fine against IWC and then broken the second TFM here silently — nobody developing on Windows would have noticed.

Both TFMs are built and checked below.

## The one line that must not be tidied away

```xml
<Compile Remove="core\**" />
```

The Web SDK globs `**/*.cs`. Without this, every file in the subtree is compiled **twice** — once into `RadioWebControl.Core.dll` and once straight into this assembly — and the error names duplicate types rather than the cause. `Content`, `None` and `EmbeddedResource` are removed alongside it, matching how `Workers/` is already excluded a few lines up.

## Verified

- `dotnet build -f net10.0-windows` — clean, 0 warnings.
- `dotnet build -f net10.0` — clean, 0 warnings.
- The **exact CI publish command** (`--framework net10.0-windows --configuration Release --runtime win-x64 --self-contained true --output publish`) lands `RadioWebControl.Core.dll` beside `Yaesu_Web_Control.exe`.
- `installer.nsi` needs **no change** — it takes the publish folder wholesale.

## Not verified

- An actual install over an existing install.
- The macOS/Linux host **on real macOS or Linux** — only its TFM was compiled here, on Windows. @fabiojvalente, if you get a moment, a `dotnet build -f net10.0` on your Mac is the one check I can't do.
- Anything on the radio. **This change cannot reach CAT** — `DxSpot` is a cluster DTO and never touches the wire.

## Why subtree and not submodule

A plain `git clone` of this repository must still build with no extra steps. Commands are in `core/README.md`; to pull later core changes down:

```bash
git subtree pull --prefix=core https://github.com/mm5agm/Radio_Web_Control_Core.git main --squash
```

Companion PR in IWC does the identical thing there. Nothing else moves — the rest migrates **opportunistically, on touch**, not in a batch.


fabiojvalente:
Hi @mm5agm 

Will try later locally in the meantime I believe you can test this your self by triggering a build on the branch

Head to actions and trigger a new one, ti will cover windows, linux (docker) and macos 

image

The build seems to be failing 
https://github.com/mm5agm/Yaesu_Web_Control/actions/runs/31714904080

I believe the workflow/action needs to checkout the repo

mm5agm:
Hi Fabio — thanks for `feauter/enhance-subtree`. I've gone through it properly and it's genuinely useful work: the scripts guard exactly the traps that bit us on phase 1, and the docs are accurate about the thing that matters most — that `core/` is already in a clone and a subtree pull must never become a build step.

Two housekeeping points and one real bug.

### Please open a PR for it

There's no PR on that branch, so nothing has run against it. That mattered less last week, but #101 has just landed a `pull_request:` trigger on both this repo and Icom Web Control, so a PR into `develop` now gets a real Windows publish + NSIS build and a Docker build before merge — which is precisely what would have caught the `core/` restore breakage in #99. A branch push on its own still gets nothing.

Also, minor: the branch is `feauter/enhance-subtree` — typo for `feature/`. Worth renaming before the PR so it sorts with the others. It's currently 3 commits behind `develop` (#100 and #101), so a rebase at the same time would be tidy.

### The bug: `update-core.ps1` fails on its first run on Windows

Line 68:

```powershell
& git remote get-url $RemoteName *> $null
if ($LASTEXITCODE -ne 0) {
    Write-Host "Adding git remote $RemoteName -> $Url"
    ...
}
```

In **Windows PowerShell 5.1**, redirecting a native executable's stderr wraps each stderr line in an `ErrorRecord`, and with `$ErrorActionPreference = 'Stop'` at the top of the script that becomes a *terminating* error. So on the one run where the remote genuinely doesn't exist — the first run, on every fresh clone — the script dies at that line instead of falling through and adding the remote.

Reproduced on a clean clone of your branch:

```
PS> .\scripts\update-core.ps1
git.exe : error: No such remote 'radio-core'
At ...\scripts\update-core.ps1:68 char:1
+ & git remote get-url $RemoteName *> $null
    + FullyQualifiedErrorId : NativeCommandError
exit 1
```

I checked both redirect forms — `*> $null` and `2>$null` both throw. This is 5.1-specific behaviour, so if you're testing under `pwsh` 7 you won't see it — worth knowing that this box has 5.1 only, no `pwsh` installed, so here it fails every time on first use.

The fix that avoids the whole class of problem is to not redirect a native command at all:

```powershell
$remotes = & git remote
if ($LASTEXITCODE -ne 0) {
    throw 'git remote failed.'
}
if ($remotes -notcontains $RemoteName) {
    Write-Host "Adding git remote $RemoteName -> $Url"
    Invoke-Checked git @('remote', 'add', $RemoteName, $Url) "git remote add $RemoteName"
}
```

Verified end to end on a clean clone with no `radio-core` remote:

```
Adding git remote radio-core -> https://github.com/mm5agm/Radio_Web_Control_Core.git
git subtree pull --prefix=core radio-core main --squash
Subtree is already at commit ea28c6c6f33c23ac08e2e5994113b0ae6e1d490b.
```

Line 44 (`& git rev-parse --is-inside-work-tree 2>$null`) is the same trap on a colder path — run outside a git repo it will throw `NativeCommandError` rather than print your friendly `not a git repository`. Same one-line style of fix if you want it.

The `.sh` is **not** affected — `if ! git remote get-url … >/dev/null 2>&1` is correct, and I confirmed it adds the remote and pulls cleanly on a first run.

### Everything else I tested, and it all passed

- `update-core.sh`: dirty-tree guard, missing `core/` guard (the error helpfully carries the first-time `subtree add` line), unknown-action guard, `--help`, first-run remote add, and the pull itself.
- `update-core.ps1` beyond line 68: dirty-tree guard, and the pull once the remote exists.
- The `git-subtree` availability check resolves correctly against Git for Windows (`.../mingw64/libexec/git-core/git-subtree`).
- `Makefile` — targets correct, and I'm glad you made the point in the comment that they're deliberately not dependencies of `build`/`run`.
- `core/README.md` documents the same operations with the URL inline rather than a named remote; the two don't contradict each other, so no change needed.

One worry I had going in, now retracted: that `push-core` would diverge, because `subtree pull --squash` flattens history and a later `subtree push` might not find a common base. Rather than guess, I tested it — `git subtree split --prefix=core` on a fresh clone resolves to `ea28c6c`, identical to `radio-core/main`, sharing a merge base with identical trees. Your `push-core` target is sound.

### One suggestion

The same subtree, and the same absence of these scripts, exists in [Icom Web Control](https://github.com/mm5agm/Icom_Web_Control). If you're happy to, the matching pair there would be welcome — the two repos are meant to stay symmetrical around `core/`.

Thanks again — this is exactly the kind of thing that stops the subtree becoming folklore.

PR #106, opened 2026-08-14
Consume the shared AdifParser and calibration-engine

mm5agm:
Consume `AdifParser` from the shared core (Radio_Web_Control_Core), dropping the local copy.

- Subtree-pulled the core's AdifParser + first test project into `core/`.
- Deleted the app's own `Services/AdifParser.cs` (150 lines, was byte-identical to the core version bar branding).
- `MemoryController` now imports `RadioWebControl.Core.Services` and calls the renamed `AdifModeToRadioMode`.

Builds 0 warnings / 0 errors. Depends on the core PR (mm5agm/Radio_Web_Control_Core).


mm5agm:
Follow-up on this branch: consume the shared calibration-engine.js.

The engine now lives in the core subtree (`core/js/calibration/`). A
`CopySharedCoreJs` MSBuild target copies `core/js` into `wwwroot` before
AssignTargetPaths, so the file is first-class Content on a clean first build
(EnableDefaultContentItems is false and the wwwroot glob is evaluated before
targets run). The wwwroot copy is git-ignored -- core/js is the single source
of truth. Import paths are unchanged.

Verified locally: clean first build regenerates the file and lists it in the
static web assets manifest; a second build adds no duplicate Content item;
publish lands it in publish/wwwroot.

PR #107, opened 2026-08-14
Release the FTdx101 front-panel meters outside of transmit

mm5agm:
## The problem

`MeterPollingService` sends `MS13;` (left = Compression, right = SWR) on **every 500 ms poll cycle**. That is what makes the FTdx101 touchscreen meters unchangeable - whatever the operator picks is stomped twice a second - and it is the only thing a video stream of the front panel can ever show.

## The key observation

Both borrowed values are **TX-only**. `MeterPollingService` already forces `SWRMeter = 0` and `CompressionMeter = 0` whenever the radio is not transmitting. So during receive the poller was pinning the display to harvest readings it then threw away.

Every other meter is read with a command that does not care what is on the screen: `RM5` power, `RM4` ALC, `RM7` IDD, `RM8` VDD, `RM9` temp, `SM0`/`SM1` S-meter. Only compression and SWR need the meter slots, and only during transmit.

So no duty-cycle sampling is needed. Borrow at TX, hand back afterwards, and the radio shows the operator's own meters for the whole of receive.

## What this PR does

1. **`MS13` only while transmitting.** On TX start, borrow the pair and read `RM0` as before. On TX end, hand them back.
2. **Return delay of 10 s.** Deliberately longer than a between-overs gap. Releasing the instant TX drops would flap the display once per over on SSB and strobe it on QSK CW break-in, which would be worse than leaving it pinned. A normal QSO keeps the meters borrowed for its duration; they come back during genuine idle time, which is when a front-panel video feed is worth watching anyway.
3. **Learn the operator's selection.** `MS` is Set/Read/Answer/**AI** on every supported model, and `GetInitialValues` already sends `MS;` - the answer was just dropped on the floor. Added an `MS` case to `CatMessageDispatcher` that records it, plus the auto-info the radio pushes when the operator changes meters on the front panel.
4. **One settle cycle discarded** after the switch (500 ms), so the first `RM0` after `MS13` cannot report a value still belonging to the previous selection.
5. **Shutdown restore** returns the operator's own selection instead of the hard-coded `MS01`.

## Compatibility with the other Yaesu models

I checked the `MS` definition in every CAT manual in `docs/manuals/`, and **the parameter encoding is different on almost every model**:

| Model | Format | Codes |
|---|---|---|
| FTdx101MP/D | `MSP1P2;` | MAIN 0=POW 1=COMP 2=TEMP / SUB 0=ALC 1=VDD 2=ID 3=SWR |
| FTdx10 | `MSP1P2;` | 0=PO 1=COMP 2=ALC 3=VDD 4=ID 5=SWR, P2 fixed 0 |
| FT-710 | `MSP1P2;` | 0=PO 1=COMP 2=ALC 3=VDD 4=ID 5=SWR, P2 fixed 0 |
| FTDX5000MP/D | `MSP1;` | 0=COMP 1=ALC 2=PO 3=SWR 4=ID 5=VDD |
| FT-991A | `MSP1;` | 0=COMP 1=ALC 2=PO 3=SWR 4=ID 5=VDD |
| FTDX3000 | `MSP1;` | 0=COMP 1=ALC 3=SWR 4=ID 5=VDD (no PO) |

`MS13` therefore means completely different things depending on the radio, and would be malformed on the single-parameter models.

Two things keep this safe:

- **Only FTdx101MP/D ever write `MS`.** That was already true (the `useRm0Pair` gate) and is unchanged. FTdx10 / FT-710 / FTDX3000 / FTDX5000 / FT-991A read compression and SWR with `RM3`/`RM6` directly and never touch `MS`, so **their front-panel meters were always free** and no CAT traffic changes for them at all.
- **The captured selection is stored as raw digits and never interpreted.** We replay exactly what the radio told us it had, which works on 1-digit and 2-digit models alike with no per-model table. The new dispatcher case is read-only and format-agnostic.

## For Fabio (@fabiojvalente) - video stream

This is the change that lets the front panel in the video feed show the operator's own meters instead of being permanently stuck on Compression + SWR. Worth testing together with the audio/video work to confirm:

- the meters visibly return to the operator's choice about 10 s after an over
- the flip at TX start is not distracting on the stream
- nothing in the stream pipeline is upset by the meter area changing

If 10 s is wrong for streaming, `MeterReturnDelay` in `MeterPollingService` is a one-line change and is the obvious candidate to promote to a Settings field.

## Bench testing needed (FTdx101MP)

- [ ] Change the meters on the touchscreen during receive - do they stay changed?
- [ ] Transmit - do they flip to Comp + SWR, and does SWR still read correctly in the web UI?
- [ ] Does SWR come up as fast as before, or is it a beat late? (If late, `BorrowSettleCycles` is the knob.)
- [ ] After ~10 s of receive, do the operator's meters come back?
- [ ] Quick CW / QSK overs - is the display stable rather than flapping?
- [ ] Quit YWC - are the operator's meters restored (not `MS01`)?
- [ ] Confirm the log shows `[Meters] Operator front-panel meter selection is MS..` at startup, which proves the `MS;` read is being captured.
- [ ] Any other model (FTdx10 in particular) - confirm compression and SWR are unchanged.

## Deliberately not in this PR

**Pre-arming.** When YWC itself initiates TX (web UI, voice, rigctld) we could borrow the meters before the radio keys, removing the settle delay for those paths. Hardware PTT (mic or foot switch) cannot be pre-armed, so it would only help some cases, and it adds cross-service coupling. Left as a follow-up if the settle cycle proves to be a problem in testing.

**Releasing the left meter slot.** We only strictly need SWR, which lives in the right slot. If `RM3` reads compression standalone on the FTdx101 (it does on every other model), we could borrow only the right slot and leave the operator's left meter alone even during transmit. That needs a bench measurement on real hardware before it can be relied on - happy to do it as a follow-up if you want to check whether `RM3` returns sane compression values on the MP.

No version bump or release notes in this PR.

### DIFF
--- a/Services/CatCommands.cs
+++ b/Services/CatCommands.cs
@@ -21,8 +21,14 @@
         // METER READING COMMANDS (RM)
         public const string MeterPower = "RM5";    // Power output meter (0-255)
         public const string MeterSWR = "RM6";      // SWR meter (0-255) — NOTE: RM6 returns stale/wrong values on FTdx101MP; use SetMetersSWR+MeterBoth instead
-        public const string SetMetersCompAndSWR = "MS13"; // Select Compression(left) + SWR(right) for RM0 read
-        public const string SetMeterPower       = "MS01"; // Restore the default Power meter on shutdown so the user's radio is left in a normal state
+        public const string SetMetersCompAndSWR = "MS13"; // FTdx101MP/D ONLY: Compression(left) + SWR(right) for the RM0 read
+        public const string SetMeterPower       = "MS01"; // FTdx101MP/D ONLY: POW(left) + ALC(right) — the radio's own default pair
+
+        // Fallback used when restoring the operator's meters and YWC never saw an
+        // MS answer (radio powered up mid-session, init answer lost, etc.). MS01 on
+        // the FTdx101 is POW + ALC, which is what the radio shows out of the box —
+        // the same value the pre-existing shutdown restore has always used.
+        public const string DefaultFtdx101MeterSelection = "01";
         public const string MeterBoth = "RM0";            // Read both currently-selected meters: RM0LLLRRR;
         public const string MeterALC = "RM4";      // ALC meter (0-255)
         public const string MeterComp = "RM3";     // Compression meter (0-255)
--- a/Services/CatCommands.cs
+++ b/Services/CatCommands.cs
@@ -22,13 +22,16 @@
         public const string MeterPower = "RM5";    // Power output meter (0-255)
         public const string MeterSWR = "RM6";      // SWR meter (0-255) — NOTE: RM6 returns stale/wrong values on FTdx101MP; use SetMetersSWR+MeterBoth instead
         public const string SetMetersCompAndSWR = "MS13"; // FTdx101MP/D ONLY: Compression(left) + SWR(right) for the RM0 read
-        public const string SetMeterPower       = "MS01"; // FTdx101MP/D ONLY: POW(left) + ALC(right) — the radio's own default pair
+        public const string SetMeterPower       = "MS00"; // FTdx101MP/D ONLY: POW(MAIN) + ALC(SUB) — the radio's own default pair
 
         // Fallback used when restoring the operator's meters and YWC never saw an
-        // MS answer (radio powered up mid-session, init answer lost, etc.). MS01 on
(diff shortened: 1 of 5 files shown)


fabiojvalente:
@mm5agm 

I've tested but I don't think this affected the mighty ftdx10 as it only have one meter (you can configure it for the ALC etc) and it stays as configured during and after TX

Before
image

TX
image


After
image

PR #108, opened 2026-08-15
Experiment: show only the meters that are actually reading

mm5agm:
> **Scope narrowed since this was opened.** The S-meter transmit fix that was
> originally part of this branch has been split out and merged to `develop` on
> its own (`264e108`), because it is a defect fix rather than a preference and
> should not wait on the question below. It is described under "Related" at the
> end - worth reading, because one of the test questions still concerns it.
> This PR is now **only** the dim-vs-hide experiment and the row reorder.

## What this is

Two separable changes to the meter row.

An experimental toolbar button, "Meters:", cycles three modes. It is there so
the two candidate behaviours can be compared against a real rig. **One of the
two is meant to be deleted before this ships.**

| Mode | Behaviour |
|---|---|
| `all` | Current behaviour, unchanged. The default. |
| `dim` | Transmit-only gauges greyed in place while receiving. Nothing moves. |
| `hide` | Transmit-only gauges removed from layout while receiving. The row reflows. |

The mode persists in `localStorage` under `ywc.meterIdleMode`.

Which gauges count as transmit-only is declared in the markup, not in the JS: a
cell carries `data-meter-scope="tx"` if it means nothing while receiving. That
is SWR, Power, Compression, ALC and IDD - all five are already pinned to zero
during receive by `MeterPollingService`, so nothing live is being suppressed.
Temp and VDD are deliberately NOT tagged: temperature is polled outside the
`isTransmitting` branch, and a hot PA or a sagging supply is worth seeing on
receive, which is exactly when you would look.

The second change reorders the row so the always-meaningful gauges are
contiguous on the left and the transmit-only block sits last:

```
S-A  S-B  Temp  VDD  |  SWR  Power  Comp  ALC  IDD
```

That matters to the `hide` variant specifically: with the transmit-only gauges
last and contiguous, hiding them removes a trailing block, so nothing that
survives changes position.

## Testing on an FTdx10 (@fabiojvalente)

Worth knowing before you look: the FTdx10 renders a much smaller meter row than
the FTdx101MP, because Temp, VDD and IDD are FTdx101/FTDX3000-only and there is
no SUB receiver.

|  | FTdx101MP | FTdx10 |
|---|---|---|
| Receive gauges | S-A, S-B, Temp, VDD | S-A only |
| Transmit gauges | SWR, Power, Comp, ALC, IDD | SWR, Power, Comp, ALC |

So in `hide` mode on receive your row collapses from five gauges to one, where
mine goes from nine to four. That is the most extreme case this change
produces, which is exactly why your opinion is the useful one.

What would help:

1. **Does `hide` look right or look broken on receive?** One gauge on an
   otherwise empty row is a big change. Blunt reaction is what is wanted.
2. **Does `dim` read better than `hide` on your rig?** On the FTdx101MP the
   preference was hide, but that was with four gauges surviving, not one.
3. **Does the FTdx10 freeze its S-meter during transmit too?** See "Related"
   below for what the FTdx101MP does. On develop the S-meter is now greyed and
   zeroed for the duration of an over. If the FTdx10 genuinely keeps measuring
   while transmitting, that would be wrong for your radio and I would want to
   know - the measurement is from one radio and I would not assume it
   generalises.
4. **Anything ugly in the reflow** when switching RX to TX repeatedly in
   `hide` mode - flicker, gauges redrawing at the wrong size, layout jumping.
   `hide` uses `display:none`, which collapses each gauge and makes its
   `ResizeObserver` rebuild it on reappearance. That happens every over, and
   it is a known cost of the hide variant that `dim` does not have.

No need to review the row reorder - it only moves Temp and VDD, which your
radio does not render, so it is invisible on an FTdx10. It does need review
from an FTdx101MP or FTDX3000 user before it ships, since it changes gauge
order for everyone.

## Related: the S-meter transmit fix (already on develop, `264e108`)

While checking whether the S-meters should also be suppressed on transmit, I
measured what the radio actually reports. With YWC stopped, I sampled `TX;`,
`SM0;` and `SM1;` together at about 21 Hz over the serial port on an FTdx101MP:

```
RECEIVE   n=375   SM0 mean 154.2, 29 distinct   SM1 mean 72.8, 21 distinct
TRANSMIT  n=101   SM0 mean 153.3,  2 distinct   SM1 mean 74.3,  2 distinct
```

101 consecutive transmit samples returned `SM0=154 SM1=75` unchanged. The "2
distinct" is the single unkey-transition sample. The last receive value before
key-down latches, and both resume within about 200 ms of release.

`MeterPollingService` polls SM0/SM1 with no transmit gate, unlike SWR which is
forced to zero on receive. So **released YWC drew two plausible-looking needles
throughout every over that were a snapshot from key-down.** On develop they are
now greyed and zeroed for the duration. The two go together: a zeroed needle on
its own would claim "no signal", a different untruth, but greyed-and-zero reads
as "not measuring". Zeroing is applied at the top of `updateSMeter`, so the
gauge, the S-unit label, the raw readout and the 30-second history strip all
agree.

Incidental finding worth recording: **TUNE registers as `TX2`**, so it is a
valid way to drive transmit tests without putting out a signal.

## Open questions before any of this ships

- **Whether to ship the visibility modes at all.**
- **Whether the row reorder ships independently.** It stands on its own, but it
  changes gauge order for 100+ existing users, which is muscle memory.
- When a variant wins, the losing variant and the "Meters:" toolbar button come
  out together. The `data-meter-scope` tagging is the part worth keeping, and
  is a candidate for the shared core so IWC gets it too.

### DIFF
--- a/Pages/Index.cshtml
+++ b/Pages/Index.cshtml
@@ -228,6 +228,23 @@
         display: none !important;
     }
 
+    /* The mirror case, and a real defect in its own right: the radio FREEZES
+       both S-meters at key-down. Measured on the MP 2026-08-15 — 101 samples
+       across a 5 s over returned one value each (SM0=154, SM1=75), against 29
+       and 21 distinct values while receiving. The needles look alive but are
+       showing whatever the band was doing at the instant of key-down.
+
+       Dimmed rather than hidden because the S-meters are leftmost: hiding them
+       would shift the whole row left on every over, which is exactly the
+       movement the reorder removed. Dimmed rather than zeroed because a needle
+       dropping to S0 asserts "no signal", which is a different untruth. Grey
+       reads consistently across the row as "not telling you anything now". */
+    .meters-tx-dim [data-meter-scope="rx"] {
+        filter: grayscale(1);
+        opacity: 0.3;
+        transition: opacity 0.25s ease, filter 0.25s ease;
+    }
+
     /* Voice command help popup (right-click a mic button). */
     #voiceHelpBody .vh-intro   { color: #bbb; font-size: 0.85rem; margin-bottom: 0.75rem; }
     #voiceHelpBody .vh-group   { margin-bottom: 0.85rem; }
@@ -524,7 +541,7 @@
                         <!-- S-Meter A (MAIN) — moved from per-VFO panels to the top
                              row in v2.3.9. Always live (unlike SWR/Power/etc which
                              are TX-only). -->
-                        <div class="text-center" style="position: relative; height: 135px; min-width: 165px; margin: 0; padding: 0;">
+                        <div class="text-center" data-meter-scope="rx" style="position: relative; height: 135px; min-width: 165px; margin: 0; padding: 0;">
                             <div class="analog-meter-container">
                                 <canvas id="sMeterCanvas" role="img" aria-hidden="true" aria-label="S meter A" data-a11y-key="meters.smeterA"></canvas>
                             </div>
@@ -541,7 +558,7 @@
                                         data-a11y-key="meters.smeterHistoryB"
                                         style="display:block; width:100%; height:100%; border-radius:3px;"></canvas>
                             </div>
-                            <div class="text-center" style="position: relative; height: 135px; min-width: 165px; margin: 0; padding: 0;">
+                            <div class="text-center" data-meter-scope="rx" style="position: relative; height: 135px; min-width: 165px; margin: 0; padding: 0;">
                                 <div class="analog-meter-container">
                                     <canvas id="sMeterCanvasB" role="img" aria-hidden="true" aria-label="S meter B" data-a11y-key="meters.smeterB"></canvas>
                                 </div>
--- a/Pages/Index.cshtml
+++ b/Pages/Index.cshtml
@@ -206,6 +206,28 @@
         display: none !important;
     }
 
+    /* EXPERIMENT (feat/meter-visibility-experiment) — de-emphasise the
+       transmit-only gauges while receiving. Two candidate behaviours, cycled
+       from the "Meters:" toolbar button; one gets deleted before this ships.
+       See wwwroot/js/ui/meter-visibility.js for which gauges are tagged and
+       why Temp and VDD are deliberately NOT among them.
+
+       dim: geometry untouched, so nothing reflows when TX/RX flips mid-QSO.
+       Costs nothing to the layout, buys no width back either. */
+    .meters-idle-dim [data-meter-scope="tx"] {
+        filter: grayscale(1);
+        opacity: 0.3;
+        transition: opacity 0.25s ease, filter 0.25s ease;
+    }
+
+    /* hide: reclaims the row, at the price of a reflow on every over. Note
+       this does NOT reliably fix the narrow-viewport wrap warning below —
+       transmit still needs room for the full set, so a row that wraps at
+       nine may well still wrap at seven. */
+    .meters-idle-hide [data-meter-scope="tx"] {
+        display: none !important;
+    }
+
     /* Voice command help popup (right-click a mic button). */
     #voiceHelpBody .vh-intro   { color: #bbb; font-size: 0.85rem; margin-bottom: 0.75rem; }
     #voiceHelpBody .vh-group   { margin-bottom: 0.85rem; }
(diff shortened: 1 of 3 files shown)


fabiojvalente:
Haven't tested it yet, but why not let the user define what meters to show ?
From my findings the metrics are there/available regardless of what the user configures in the ui, so I'd let the user drive this

Will test tomorrow/coming days

fabiojvalente:
> Does hide look right or look broken on receive? One gauge on an
otherwise empty row is a big change. Blunt reaction is what is wanted.

The S-meter is the only on RX, during TX the remaining gauges appear and S-meter goes gray (still shows)
After TX it goes back to show only the S-Meter

> Does dim read better than hide on your rig? On the FTdx101MP the
preference was hide, but that was with four gauges surviving, not one.

To me, a not so **_affected by eyesight problems_** operator it makes little to no difference, at most it changes my focus or where to look, but I get the point it may help the ones needing some hints on where to focus :)

It works as expected

> Does the FTdx10 freeze its S-meter during transmit too? See "Related"
below for what the FTdx101MP does. On develop the S-meter is now greyed and
zeroed for the duration of an over. If the FTdx10 genuinely keeps measuring
while transmitting, that would be wrong for your radio and I would want to
know - the measurement is from one radio and I would not assume it
generalises.

No. The FTDX10 does not measure S units during TX, instead it measures the secondary display, which is one of P0, COMP, ALC, VDD, ID, SWR

From my findings all of them are available (over CAT) at TX, the radio meter just shows one of them - and YWC doesn't show all of them either - worth following up on a different PR that I never bothered to work on as video-tx has been taking priority over this one

> Anything ugly in the reflow when switching RX to TX repeatedly in
hide mode - flicker, gauges redrawing at the wrong size, layout jumping.
hide uses display:none, which collapses each gauge and makes its
ResizeObserver rebuild it on reappearance. That happens every over, and
it is a known cost of the hide variant that dim does not have.

Nothing really ugly, all resizes as expected

I'm just wondering if handling this behaviour (dim/enable/hide show etc( can bite us back in the future while trying to cover all the usecases, other than that am all good 👍

Also, see my previous comment

fabiojvalente (approved):

mm5agm:
Thanks for testing this properly and for the approval, @fabiojvalente. I am going to
park it rather than merge it, and your two comments are the reason - so let me set out
the reasoning rather than just letting it go quiet.

## First, the most valuable thing in your reply

The FTdx10 not measuring S units during transmit either. I had one measurement, from one
radio, and I was not willing to assume it generalised. Your answer means the fix already
on develop - greying and zeroing the S-meters for the duration of an over - is correct
for your radio as well as mine, rather than something I imposed on it. That was the part
I most needed and it is already shipped, independently of this PR.

## Why park

Your two points are really one point, and I think you are right.

"Let the user define what meters to show" and "this can bite us back while trying to
cover all the use cases" are the same objection: the automatic receive/transmit mode may
be the wrong abstraction. What decided it for me is that the two are not interchangeable.
A static picker cannot express "show Power during the over and not on receive". The
automatic mode cannot express "I never want to see IDD at all". So if I merge this now
and add the picker later, YWC owns both mechanisms permanently, plus an explanation of
how they interact. That is exactly the bite you are describing, and it is far cheaper to
avoid than to unwind afterwards.

The stronger reason is your other finding. If all six transmit meters are readable over
CAT while the radio only displays one, and YWC currently shows fewer than it could, then
the set of gauges on that row is likely to change. Deciding visibility policy and gauge
order for today's nine gauges, when the count is about to move, is settling the question
against a layout that will not exist. Capability first, presentation afterwards, in that
order.

So the separate PR you floated is the one I would rather see first. I would be glad if
you picked it up whenever video-tx frees you, and equally happy to do it myself if you
would rather not - your finding either way, and it is the more interesting change of the
two. There is no rush from my side.

## What parks with it, and why

The row reorder goes too, which is worth being explicit about because it looks like it
could ship on its own. It cannot, usefully. Grouping the transmit-only gauges last is
what made hide viable in the first place - the hidden block is trailing, so nothing that
survives moves. On its own it changes gauge order for everyone already running YWC and
delivers nothing visible in return. That is the worst of both, so it waits.

## What would bring it back

Evidence that the reduced clutter genuinely helps a partially sighted operator. You were
straight about not being affected yourself, and that is useful to know rather than a
gap - it tells me the change is not broken, which is what I asked you for, but it cannot
tell me whether it is wanted. Until one of the operators it was aimed at asks for
something like it, parked is the honest place for it to sit.

Converting to draft rather than closing, so your review and this discussion stay attached
to it and the branch is still there if the answer comes back yes.

73 Colin MM5AGM

mm5agm:
Closing as a parked experiment rather than a rejected idea.

It has been a draft since 16 August and develop has moved a long way under
it since. If the meter-visibility question comes back, it is cheaper to
redo it against current develop than to revive this branch - so nothing
here is worth keeping open on the list.

The branch is not deleted, so the diff is still there if it is wanted.

PR #109, opened 2026-08-15
Design note: drive the radio's own scope and meters over CAT

mm5agm:
Documentation only. No code changes, nothing shipped, no behaviour altered.

## The idea

With USB video capture we can see the radio's own TFT in the browser but we
cannot touch it. The obvious fix was to draw clickable overlays on top of the
video, positioned over the radio's on-screen buttons. That means knowing where
every button sits in the captured frame -- which varies by model, by firmware
screen layout, by capture resolution, and by what the operator currently has
displayed. It is pixel geometry pretending to be a UI.

The alternative: don't touch the radio's screen, command the radio. Yaesu
exposes the scope and meter controls over CAT. Send the command, the radio
changes its own display, and the capture shows the result because it is showing
the radio's actual screen.

Two things worth being clear about:

- This does not replace the video capture work. You still need the capture to
  see the screen. It removes the overlay/compositing layer specifically, which
  was always going to be the fragile part.
- It is worth having without video at all. Remote scope control stands on its
  own for anyone operating from another room, which reaches every user with a
  supported radio rather than only those with a capture dongle.

## What the radio actually answered

Added `scripts/probe/ss-probe.ps1` -- a read-only probe. It sends Read frames
only; no Set frame was sent, so nothing on the radio changed. Run against a
real FTdx101MP (ID0682) on COM4 with YWC stopped.

All 18 reads answered. Highlights:

```
SS04;  LEVEL  -> 'SS04+05.0;'
SS05;  SPAN   -> 'SS0590000;'     MAIN: 1 MHz
SS15;  SPAN   -> 'SS1580000;'     SUB:  500 kHz
SS06;  MODE   -> 'SS0660000;'     MAIN: W/F CURSOR (L)
SS16;  MODE   -> 'SS1620000;'     SUB:  3DSS FIX
SS08;  HOLD   -> 'SS0800000;'
```

That settles two things the layout-mangled CAT manual PDF could not:

1. P3-P7 is one 5-character field, not five one-character fields. `+05.0`
   proves it. Most sub-commands use only the first character and pad with
   zeros.
2. P1 really does address MAIN and SUB independently -- MAIN was on 1 MHz W/F
   CURSOR while SUB was on 500 kHz 3DSS FIX at the same moment.

## Per-radio picture

Full matrix is in the note. The four differences that matter:

- **Mono/Multi has no CAT command on any of the seven models.** Searched every
  CAT manual; no command, no menu entry. This is the one item on the original
  wish-list that stays front-panel-only. If it is essential it needs an
  overlay -- but a much smaller one than the original proposal.
- **The FT-710 has no HOLD.** Its P2 list stops at 7; the FTdx101 and FTdx10 go
  to 8. Genuinely fewer functions, not a documentation gap.
- **The FT-710 spells Expand differently** -- EXPAND/NORMAL where the 101 and
  10 use L/N/S. Incompatible value tables, so the mode selector needs a
  per-model map. Do not unify them.
- **FTDX3000 and FTDX5000 are effectively out.** FTDX3000 has no SS at all
  (scope lives in EX menu 124-148, which is configuration rather than a
  control); FTDX5000 has only a DP display-page selector. Neither has an
  engaged reporter.

## What is NOT established

No write has been attempted. Every frame in the note is a Read. The first
bench gate before any code is to send one Set by hand and read it back --
`SS0540000;` to move MAIN span to 20 kHz, then `SS05;` to confirm it reads
back 4.

Also open: whether SS frames arrive over auto-information when the operator
touches the radio's own screen (determines sync vs poll), and how a
user-facing meter selector would cooperate with the existing borrow-and-restore
in MeterPollingService, which overwrites the meter pair on transmit. That
interaction is the most likely bug in the whole feature.

The FTdx10 and FT-710 columns are from the manuals only. @fabiojvalente -- if
this looks worth doing, the FTdx10 column is the one you could confirm, though
there is nothing to test yet and no rush.

## Scope recommendation

If it goes ahead: FTdx101 and FTdx10 only, per the focus-on-actively-reported-
radios rule. FT-710 support is cheap to add later but would ship unverified.

### DIFF
--- a/Controllers/ScopeController.cs
+++ b/Controllers/ScopeController.cs
@@ -1,0 +1,356 @@
+using Microsoft.AspNetCore.Mvc;
+using Yaesu_Web_Control.Services;
+
+namespace Yaesu_Web_Control.Controllers;
+
+/// <summary>
+/// Drives the radio's OWN spectrum scope over CAT (the SS command).
+///
+/// This is not YWC's SDR spectrum panel — that is separate hardware fed by an
+/// RSP1 on the rear-panel IF output, and nothing here touches it. What these
+/// endpoints change is what the operator sees on the radio's front-panel TFT,
+/// which is useful when operating remotely, and useful again when that TFT is
+/// being captured to the browser as video. See
+/// docs/design/scope-control-via-cat.md for why this beats drawing clickable
+/// overlays on top of the captured picture.
+///
+/// Deliberately a separate controller rather than more endpoints bolted onto
+/// CatController (which is already 3000 lines): this feature is new, is gated
+/// per model, and may not survive contact with users. Keeping it in one file
+/// means it can be judged — or deleted — on its own.
+///
+/// Every write is followed by a read-back, and the read-back is what the API
+/// returns. The radio, not our optimism, is the source of truth for what the
+/// scope is now showing; that matters here more than usual because the operator
+/// may be sitting in front of the radio changing the same settings by hand.
+/// </summary>
+[ApiController]
+[Route("api/scope")]
+public class ScopeController : ControllerBase
+{
+    private readonly ICatClient _catClient;
+    private readonly ISettingsService _settingsService;
+    private readonly RadioStateService _radioState;
+    private readonly ILogger<ScopeController> _logger;
+
+    // SS frames are short and the radio answers them promptly, but they share
+    // the port with the ~10 Hz meter poll. One at a time, same as CatController.
+    private static readonly SemaphoreSlim _gate = new(1, 1);
+
+    // How long to let the radio redraw before reading a setting back after a
+    // write. Measured on an FTdx101MP: without it, roughly one read in three
+    // following a mode change goes unanswered.
+    private const int SettleAfterWriteMs = 150;
+
+    // Gap before a retried read. Deliberately shorter than the settle delay —
+    // by this point the radio has already had the settle window.
+    private const int RetryDelayMs = 80;
+
+    public ScopeController(
+        ICatClient catClient,
+        ISettingsService settingsService,
+        RadioStateService radioState,
+        ILogger<ScopeController> logger)
+    {
+        _catClient       = catClient;
+        _settingsService = settingsService;
+        _radioState      = radioState;
+        _logger          = logger;
+    }
+
+    // ── Endpoints ────────────────────────────────────────────────────────────
+
+    /// <summary>
+    /// GET /api/scope/main | /api/scope/sub
+    /// Reads the scope settings the UI exposes, straight from the radio.
+    /// </summary>
+    [HttpGet("{band}")]
+    public async Task<IActionResult> GetState(string band)
+    {
+        if (!TryBand(band, out var p1, out var bandError))
+            return BadRequest(new { error = bandError });
+
+        if (!Supported(out var model))
+            return NotSupported(model);
+
+        if (!await _gate.WaitAsync(2000))
+            return StatusCode(503, new { error = "Radio busy" });
+        try
+        {
+            await EnsureConnectedAsync();
+            return Ok(await ReadStateAsync(p1, model));
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error reading scope state for {Band}", band);
+            return StatusCode(500, new { error = "Failed to read scope state" });
+        }
+        finally { _gate.Release(); }
+    }
+
+    /// <summary>
+    /// POST /api/scope/{main|sub}/{setting}   body: { "value": "..." }
+    ///
+    /// setting is one of: span, mode, hold, marker, peak, speed, level.
+    /// Returns the re-read state, so the caller can see what the radio actually
+    /// did rather than what it was asked to do.
+    ///
+    /// Route token is {setting}, NOT {action}: "action" is a reserved ASP.NET
+    /// Core routing token and an attribute route using it silently fails to
+    /// register, 404ing every request. That already cost a debugging session
+    /// once on the QMB endpoints — see CatController.Qmb.
+    /// </summary>
+    [HttpPost("{band}/{setting}")]
+    public async Task<IActionResult> Set(string band, string setting, [FromBody] ScopeSetRequest? request)
+    {
+        if (!TryBand(band, out var p1, out var bandError))
+            return BadRequest(new { error = bandError });
+
+        if (!Supported(out var model))
+            return NotSupported(model);
+
+        var value = request?.Value?.Trim() ?? "";
+        if (value.Length == 0)
+            return BadRequest(new { error = "value is required" });
+
+        string frame;
+        switch (setting.ToLowerInvariant())
+        {
+            case "span":
+                if (!TryDigit(value, 0, 9, out var span))
+                    return BadRequest(new { error = "span must be 0-9 (0 = 1 kHz … 9 = 1 MHz)" });
+                frame = ScopeCommands.Set(p1, ScopeCommands.Span, span);
+                break;
+
+            case "mode":
+                // The UI sends the composed P3 character rather than the three
+                // axes, so the composition rule lives in exactly one place
+                // (ScopeCommands.ModeValue, which the browser mirrors). Validate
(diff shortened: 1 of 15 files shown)


mm5agm:
@fabiojvalente - this is the idea I mentioned, and it is now working on my radio, so I would
value your opinion on it before I take it any further.

Instead of drawing clickable overlays on top of the captured TFT image, YWC sends the
radio's own SS (spectrum scope) CAT command and lets the radio change its display. The
video capture then shows the change for free. It is not a replacement for your capture
work at all - the capture is still what puts the scope on the screen. What it does
replace is the overlay and hit-target layer on top of it, which is the fiddly part, and
it also means people with no capture hardware get scope control anyway.

Branch is `docs/scope-control-via-cat` (PR #109). Panel is a partial view sitting above
the spectrum panels for now, collapsed by default.

What it drives: span, display type (W/F or 3DSS), placement (Center / Cursor / Fix),
size, marker, reference level, and Hold. Every write is followed by a read-back and the
read-back is what repaints the buttons, so if you are standing at the rig changing the
same things by hand the panel follows you rather than fighting you.

Where I have got to:

* FTdx101MP - hardware confirmed. Frame format verified with a serial probe, then every
  control exercised on both MAIN and SUB. The MAIN/SUB buttons also switch the radio's
  operating band (VS), because the radio only ever displays the scope of the band it is
  operating on.

* FTdx10 - enabled from the CAT manual only. I have never run a single one of these
  commands against an FTdx10. This is the bit I would really like you to try, if you are
  willing. Being single receiver it should render no MAIN/SUB row at all and always talk
  to scope 0. Per the manual it has Hold and the same L/N/S size labels as the '101.

* FT-710 - manual only as well, and no active tester. Its SS table is genuinely different
  (no Hold, EXPAND/NORMAL instead of L/N/S), so it is handled separately rather than
  lumped in.

Two things I found on the bench that will look like bugs and are not:

1. The radio stores a separate span for each display mode. Switch between W/F and 3DSS
   and the highlighted span moves on its own. That is the radio, and re-sending the old
   span to "correct" it would overwrite something the operator chose at the front panel.

2. A read issued immediately after a write sometimes goes unanswered while the radio
   redraws, most often after a mode change. There is a short settle plus one retry. If it
   still gets nothing it reports unknown rather than guessing.

One gap worth knowing about: Mono / Multi has no CAT command on any model. I searched all
seven manuals. If any overlay survives this, that is the one.

If you do get a chance to point it at the FTdx10, what I most want to know is whether the
panel comes up at all, whether the buttons land, and whether the read-back matches what
the radio is actually showing. No rush, and no problem at all if you would rather not.

73 Colin MM5AGM

mm5agm:
Follow-up, and I think this is the part that makes the CAT approach worth having.

I did not know whether the radio announces scope changes made at its own front panel.
It turns out it does. I put a temporary log line on the SS opcode, turned the knobs on
the rig, and got this:

    SS0510000    MAIN  span 5  (50k)
    SS0590000    MAIN  span 9  (1M)
    SS0580000    MAIN  span 8  (500k)   \ 12 ms apart
    SS0620000    MAIN  mode 2  3DSS Fix /
    SS0810000    MAIN  hold ON
    SS0800000    MAIN  hold OFF

So SPAN, MODE and HOLD all report, tagged with the band. Note the pair 12 ms apart: a
mode change arrives as two messages, the new mode plus the span that mode carries,
because the radio stores span per display mode. That means following it live gets the
per-mode-span behaviour right without any special handling.

Two honest gaps. I never saw MARKER or LEVEL come through, and I cannot tell whether
that is because the radio does not report them or because I did not happen to change
those two during the test. The handler routes on the sub-command number rather than
listing them, so it copes either way. And the radio does not echo changes YWC itself
made over CAT, only front-panel ones - which is fine, since the write's own read-back
already repaints those.

I have built the live sync on top of that: an SS message off the wire patches the one
setting that changed and repaints, with no re-read, since the radio has just told us
the value. The browser half is verified with a headless-Chrome probe. The end-to-end
test with my hand on the rig is still to come, so treat it as working-but-unconfirmed
until I say otherwise.

Why it matters to you: it means the web panel and your captured TFT image stay in step
no matter which end drives them. Change it in the browser and the capture shows it;
change it on the rig and the browser follows. That is the bit an overlay could never
do on its own.

One more thing to watch if you do try it on the FTdx10: I only know the FTdx101 does
this. Whether the FTdx10 announces SS the same way is another unknown, and an easy one
to check - open the panel, turn the span knob on the radio, see if the panel moves.

73 Colin MM5AGM

mm5agm:
@fabiojvalente - I said I would value your opinion before taking this any
further, so I owe you a note rather than just a merge notification.

I am going to land it, but only the half I have actually run against a radio,
and I want to be straight about which half that is and why I am not sitting on
it waiting.

## What is being gated off, and why

`SupportsSpectrumScopeCat` originally enabled the FTdx101, the FTdx10 and the
FT-710. The last two were enabled purely on the strength of their CAT manuals.
I have narrowed it to the FTdx101 pair.

The reason is that these are writes. `SS` does not read the scope, it changes
what appears on the front panel of somebody's radio, and my own bench run found
two behaviours that are in no manual I have: the span setting is stored per
display mode rather than globally, and a read issued straight after a write can
come back stale. If manual-derived writes to the FTdx10 turn out to have their
own version of that, the person who discovers it should not be an operator who
just installed a release. Least of all on your radio.

So the FTdx10 stays off until somebody runs it. The per-model tables for hold
and for the size labels keep their FTdx10 and FT-710 rows - that is deliberate
and noted in the code, because the manual reading is worth keeping and those
tables are never consulted while the gate is false. Turning a model back on is
one line plus a probe run.

## The one thing that would turn yours on

`scripts/probe/ss-write-probe.ps1` against the FTdx10, and a note of whether
span, hold and marker read back what you set. That is the whole ask. It is a
few minutes with the rig, not a code review, and there is no timetable on it -
video-tx is the more valuable work and I would rather you stayed on it.

## Nothing of yours is displaced by this

Worth saying plainly, because the two changes look adjacent. The thing this
supersedes is the clickable overlay I floated on 13 August - my own idea, my
own brainstorm, never started by anyone. Driving the radio over CAT turned out
to be the better route to the same place, so the overlay is what got dropped,
not any of your work.

Your capture work is what makes this visible at all. Changing the span over CAT
is invisible unless the radio's screen is on the operator's monitor, which is
your PR, not mine. They are complementary rather than competing, and if
anything this raises the value of what you have built.

## The rebase, honestly

This does add to your eventual rebase, so I would rather flag it than have you
find it. Both branches touch `Pages/Index.cshtml`: video-tx inserts a block at
around line 743 that runs on to 854, and this one inserts at 784, inside that
span, plus we both touch the module import block lower down.

The mitigating point is that you already owe a rebase there whatever I do. The
cache-busting change replaced every hand-kept `?v=N` on those imports with the
app version, which is a conflict your branch will hit on its own. Landing this
now means one rebase clears both, rather than you doing that work and then
having to do it again.

73 Colin MM5AGM

mm5agm:
Closing the loop on this, since I said above to treat live sync as
working-but-unconfirmed until I said otherwise. I am saying otherwise.

Tested on develop after the merge, against the '101MP, with the CAT log watched
from the other side while I turned the knob. Expanding the panel read all seven
sub-commands first time with no drops (span 4, mode 1, marker 1, peak 0,
speed 2, level +02.5, hold 0). Four span changes made at the radio's own front
panel then arrived unsolicited - SS0500000, SS0560000, SS0520000, SS0530000 -
and the highlighted span button in the browser followed every one, with no
re-read and without collapsing and reopening the panel.

Two honest limits. This exercises SPAN, which shares its path with every other
sub-command, so mode and hold are expected to work but I have not separately
proved them - and the front-panel mode change is the interesting one, because
it arrives as two frames a few ms apart. Marker and level I still have never
seen announced at all, so I cannot say whether the radio omits them.

The reason I wanted this nailed down rather than left as probably-fine: it is
the half that makes the CAT approach worth having next to your capture. Change
something in the browser and the captured screen shows it; change it at the rig
and the browser follows. Neither end has to be the master.

73 Colin MM5AGM

PR #111, opened 2026-08-15
Record that the seam-in-core decision needs Fabio

mm5agm:
Docs only — one new section in `core/docs/design/shared-core-plan.md`, so it is visible from both repos.

## Why now

A no-hardware **stub radio** for YWC came up tonight. The useful realisation is that the stub is the wrong unit to argue about: you cannot build one without first building YWC a seam, so **the seam question arrives the moment stub work starts, not after it** — and building a YWC-local seam then moving it to core later is the expensive order.

That is worth writing down before anyone starts, hence this.

## What it says

New **§3a**, filed under the existing "The blocker: YWC has no radio seam":

- **The prize is real** — a seam in core would be its first component with serious leverage (core today is `DxSpot.cs`, `AdifParser.cs`, `calibration-engine.js`), and both apps would get a stub nearly free.
- **The shape is not obvious** — IWC's seam is shaped around a single-receiver IC-7300; YWC spans the dual-receiver FTdx101, the FTdx10 and the FT-710. One seam spanning both is a decision about which behaviour wins, not a copy, and §4's "only move what already agrees" rule does not cover it.
- **It is adjacent to the monorepo proposal in §5**, which is parked by agreement — so it deserves a conversation with Fabio rather than a unilateral start.

It also records the two things that motivated it: the stub has already paid for itself in IWC (the Firefox meter-needle bug was reproduced *and* verified under it), and its absence here has a cost on record (PR #110 is pure HTTP-layer and still could not be runtime-verified).

Finally it notes the cheaper option — HTTP-layer changes need only a booting app, not a fake radio, so YWC pointed at a non-existent COM port would probably do. Flagged in the text as an untested guess rather than a finding.

## Note

This touches `core/`, so it is subtree content: **never squash-merge, never rebase across it.**

### DIFF
--- a/core/docs/design/shared-core-plan.md
+++ b/core/docs/design/shared-core-plan.md
@@ -137,7 +137,9 @@ second half only.
 
 ### 3a. When the seam question actually arrives — and who has to be in it
 
-Added 2026-08-15, after a no-hardware **stub radio** for YWC was raised.
+Raised 2026-08-15, after a no-hardware **stub radio** for YWC was proposed.
+Answered 2026-08-30 — see "What Fabio said" below, which supersedes the
+plan this section originally carried.
 
 IWC can run its whole UI with no radio attached (`IWC_USE_STUB_RADIO=1` swaps a
 canned `StubRadioController` in behind the seam). That has already paid for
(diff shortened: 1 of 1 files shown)


mm5agm:
@fabiojvalente A nudge on this one - it has been sitting a fortnight and it is
my fault for opening it as a wall of text. Here is the short version.

The PR itself is docs only: one new section in the shared-core plan. Merging it
changes no code. It is on your review queue because of the question it records,
not because of the diff.

**The question.** Both apps want a no-hardware stub radio. You cannot build one
without first giving the app a radio seam - so the seam question arrives the
moment stub work starts, not after it. IWC's seam is shaped around a single
receiver. YWC spans the dual-receiver FTdx101 as well as the FTdx10 and FT-710.
A seam in core spanning both is a decision about which behaviour wins, not a
copy, and it is adjacent to the monorepo idea we parked. That is why I have not
just started.

**What I actually need from you** is one of three answers, and any of them
unblocks me:

1. *Go ahead, shape it for both* - I do the work and you review the seam.
2. *Not yet* - I merge the doc as a record and leave the seam alone. The stub
   waits.
3. *Wrong framing* - you see it differently, and we talk before anything moves.

If you have no strong view, (2) is the cheapest and I am happy to take silence
as (2) after another week rather than keep pinging you. I would rather have
your (3) than my (2), though, if there is one in there.

No hurry on #120 from this - they are unrelated.

fabiojvalente:
Hi @mm5agm 

Thanks for the shorter version — that helped.

From my read, on the three options: I would pick (2), with a nuance.

Merge this as a record. **Do not start an IRadioController (or equivalent) in core/ yet.**

I take the seam as the wall (semantic radio above, CAT/CI-V below), not as the stub itself. The stub is one plug behind that wall, and it is a good reason to talk about a seam — it is not a reason to put the radio contract in `core/` now. `core/` is still the radio-agnostic slice (spots, ADIF, calibration maths). A shared radio type is a different decision from “move what already agrees,” and it sits next to the monorepo idea we parked.

Regarding the nuance, I am not saying NO to a no-hardware stub in YWC. If we want one, we should do a YWC-local seam first (thin semantic API, CAT behind it, stub as one implementation). Use it here until it has paid for itself. I would not treat **“promote to core/ when the two contracts look alike”** as the plan.
Icom and Yaesu will not become one radio (and there's a good reason most of the comercial solutions have different options for different vendors: ex: win4yaesu/win4icom). Frequency / mode / PTT can stay generic; Icom waterfall and Yaesu VC Tune, MAIN/SUB dual receive, and roofing filters should not be forced under a common name, and we should not grow a shared interface with a method per brand quirk. Those stay in each app. If the local seams never look alike, that is the honest outcome — `core/` keeps refusing to know what a radio is.

A missing COM port is enough to boot the app for some HTTP-only work; it is not a substitute for a stub when you need live frequency/meter behaviour.

mm5agm:
@fabiojvalente That is a much more useful answer than the option number. Taking
(2), and I am not starting anything in `core/`.

Agreed on the substance, and I think you have named the real risk better than I
did. The seam is the wall, not the stub. An interface that grows a method per
brand quirk is worse than two honest interfaces. VC Tune, MAIN/SUB dual receive,
roofing filters, your waterfall - those do not have a common name and forcing
one would be inventing a fake abstraction to make a diagram tidy.

One thing I will push back on, and only one: win4yaesu and win4icom staying
separate tells us what their vendors chose to sell, not what is correct. Two
products, two licences, two support queues - there are commercial reasons for
that split that have nothing to do with whether the code could have been shared.
But that argument is decoration; it is not what your position rests on. The
naming argument stands up entirely on its own and I am not arguing with it.

**Where I think there is a third option you have not been offered.** You wrote
that frequency / mode / PTT can stay generic. That generic set is most of what
the UI, the voice dispatcher and a stub actually need - none of them care about
roofing filters. So the option is not "one seam spanning both radios", which is
what my #3a asks for today and is exactly the thing you are rejecting. It is a
deliberately **narrow** seam: only the verbs we already agree are generic, with
every brand quirk staying in its own app behind its own local interface. `core/`
would still refuse to know what a roofing filter is.

I am not asking to do that now, and I am not asking you to agree to it now. The
order I would like is the one you described: YWC-local seam first, thin semantic
API with CAT behind it, stub as one implementation, used here until it has paid
for itself. Then there are two real seams to compare and the narrow question gets
answered by evidence instead of by either of us guessing in advance.

So the only thing I want from you before I merge this: **does the narrow version
change your answer, or is it still not yet?** Either is fine - I will merge the
doc recording whichever you say.

Two amendments I am making to #3a regardless, because your reply makes the
current text wrong:

- The "one seam spanning both" framing and the "promote to core when the
  contracts look alike" plan come out. That was the pre-answer framing.
- Your point that a missing COM port is not a substitute for a stub when you
  need live frequency or meter behaviour gets added to the closing paragraph.
  I had that flagged as an untested guess; you have now tested it.

fabiojvalente:
@mm5agm The narrow version does change the answer, in that sense only.

Still not yet for anything landing in `core` . Same order as before: YWC-local seam first, thin semantic API, CAT behind it, stub as one implementation, used here until it has paid for itself.

What I am saying yes to is the shape of a later question: if we ever share a radio type, it is only the verbs we already treat as generic (frequency, mode, PTT, and whatever else both apps genuinely share after two real seams exist). Brand quirks stay in each app. `core` still does not know what a roofing filter or a waterfall is.

That is not “one seam spanning both radios.” I would treat the narrow slice as the only candidate worth comparing, and only with evidence, not as a plan to start now.

Happy for you to merge the doc with that recorded, plus the two amendments you listed.

PR #112, opened 2026-08-18
Fix/audio popout nav crash

fabiojvalente:
## Summary

- Fix a host process native crash when Remote Audio is streaming in the pop-out window and the operator clicks the navbar quickly (Home / Settings / About / …). The log ended mid-stream with no managed exception — a PortAudio/WASAPI native death, not a normal shutdown.
- Serialise all PortAudio control-plane calls onto a dedicated `YWC-PortAudio` thread and return a cached device list while streams are open, so `/api/audio/devices` cannot touch `portaudio.dll` mid-callback. `AudioDeviceEnumerator` serves the cache when streams are open; `/api/audio/devices` reports `cachedWhileStreaming` when PortAudio is still open.
- Pop-out handoff no longer tears down and reopens WASAPI on the USB codec: when the Index WebSocket closes, PortAudio streams stay open for a 3 s grace window and the pop-out reconnect reuses them (`Reusing open audio devices (session handoff within grace window)` in the log). The 150 ms client-side delay was removed — the server grace replaces it. Grace-close and reconnect reuse are serialised on the PortAudio thread so a late queued close cannot dispose streams under a new session.
- Heartbeat fires immediately on SignalR connect and `Heartbeat()` cancels the idle-shutdown timer.
- Session teardown is scoped to the disconnecting WebSocket so a pop-out reconnect cannot have its pumps/socket aborted by the Index connection's `finally` block.
- **Operator-visible trades for stability:** browser playback jitter buffer increased from 3 to 6 frames (~30 ms → ~60 ms, about 30 ms more RX latency); server levels pump interval increased from 100 ms to 200 ms (meter update rate halved).

## Test plan

- [ ] **Rebuild and restart** the host (stop the debugger / tray app, then `dotnet run --project Yaesu_Web_Control.csproj --framework net10.0-windows`). Changes are not picked up by a running process.
- [ ] Start Remote Audio on Home, then **Pop out**. Confirm the pop-out shows streaming and Home shows *In pop-out window (streaming)*.
- [ ] Check the log on pop-out handoff: expect `Reusing open audio devices (session handoff within grace window)` — **not** a second `Audio devices open` line within ~1 s of disconnect.
- [ ] With the pop-out still streaming, click navbar items quickly (Home, Settings, About, Diagnostics, back to Home, repeat). The host must stay up; audio in the pop-out must keep playing.
- [ ] Open **Settings → Remote Audio** while the pop-out is streaming. RX/TX dropdowns still show the saved devices (cached list); no host crash.
- [ ] Stop from Home (should stop the pop-out session) and start again from Home, then pop out again. No "Audio session busy"; no host crash.
- [ ] Close the pop-out. Home returns to Idle / unlocked Start. Start audio in-page still works.
- [ ] Confirm the log no longer dies silently while audio is streaming — a crash would still be a hard stop with no `Audio WebSocket disconnected` line.


fabiojvalente:
Hi @mm5agm during the remote video development I found this bug, please have a look

fabiojvalente:
~Still investigating one more bug in this branch~

Bug fixed, ready for review

mm5agm:
Sorry for the slow reply Fabio - this one slipped past me for a week.

I have read the branch and built it. Both TFMs build clean (0 warnings, 0
errors), YaesuWebControl.Tests 30/30, core xUnit 27/27, core JS 8/8. I have not
bench-tested it yet - the test plan is all manual and needs the radio, so I will
run it here before merging.

The diagnosis looks right to me, and so does the shape of the fix: one dedicated
control thread for every PortAudio call, a cached device list while streams are
open, callbacks that cannot throw into native code, and no WASAPI teardown and
reopen on the pop-out handoff. That is four separate ways the old code could
kill the host, and you have closed all four. The grace window is a nicer answer
than the 150 ms client-side sleep it replaces.

Points below, roughly in the order I would deal with them.

1. Race between the grace-window close and a reconnect that arrives right on the
   boundary. In ScheduleDeviceClose the token is checked, and only then is the
   close queued onto the audio thread. A session connecting in that gap calls
   CancelScheduledDeviceClose (too late to stop the already-queued work), then
   CanReuseDevices sees _devicesOpen still true and logs "Reusing open audio
   devices" - and the queued close then disposes the streams underneath it. The
   session comes up, sends its open control frame, and streams nothing, with no
   error anywhere. Small window, but it is the same failure class this PR is
   fixing, and silent-dead-audio is hard to report. Could the token re-check and
   the reuse decision both run inside the Invoke on the audio thread, so they
   serialise against each other?

2. Playback callback: when the resampled length does not match frameCount and
   _cbPlayMono has to grow, EnsureCallbackBuffer allocates a fresh array and the
   just-resampled samples are lost, so that frame goes out as silence. The old
   code copied into the adjusted array, so this is a small step backwards.
   Sizing the buffer to max(outLen, frameCount) before the resample, or copying
   on grow, would fix it. Only bites when the buffer resizes, so a tick at
   startup rather than a constant fault.

3. AudioSessionManager: TryAcquire is now "busy until Release", but
   HasActiveSession still reports on socket state, and that is what
   /api/audio/status returns as sessionActive. So there is a window where the UI
   sees an idle session, offers Start, and the connect comes back busy. Making
   HasActiveSession use _connectionId too would keep the two consistent.

4. AudioController.ListDevices: the DevicesOpen branch is a 34-line copy of the
   branch below it, calling the same AudioDeviceEnumerator.ListInputs and
   ListOutputs. The comment says it stops Settings touching portaudio.dll, but
   the protection is entirely inside ListDevices where the stream count is
   checked - the controller branch adds no guard, only the cachedWhileStreaming
   flag. One code path with a bool for the flag would do the same job.

5. required_before_merge.md and video_debugging.md look like they came along
   from the video branch by accident - 252 lines, both about Radio Display, and
   required_before_merge.md is a review checklist rather than project docs. I
   would drop both from this PR.

6. The PR description mentions stopping /api/cat/status polling on non-Home
   pages, clearing stacked intervals and skipping overlapping polls. I cannot
   find that in the diff - site.js only gains the heartbeat-on-connect call. I
   think the polling work landed with #122; worth trimming the description so
   the branch and the summary agree.

7. Two user-visible changes that are not in the description and that I would
   like called out: the playback jitter buffer goes from 3 frames to 6, which is
   about 30 ms more RX latency, and the levels pump goes from 100 ms to 200 ms,
   halving the meter update rate. Both look like deliberate trades for stability
   and I have no objection - I would just rather they were stated, because
   latency is the thing operators notice on remote audio.

Minor, take or leave: in AudioDeviceEnumerator.Invoke the ManualResetEventSlim
is disposed when the 15 s timeout fires, while the queued work item can still
call done.Set() on it - the ObjectDisposedException gets swallowed by the loop,
so it is harmless today, but it is a loose end.

None of this is in the way of the core fix. Sort 1, 2 and 5, and I will bench it
here on the '101 against your test plan.

mm5agm:
Fabio - a heads-up before this merges, because some work on the CW reader
branch collides with it and one of the collisions is the silent kind.

We hit the same worklet defect from opposite ends. You found it as a click
while navigating the main window; I found it as a warble on a steady CW tone,
which turned out to be the same starved jitter buffer splicing discontinuities
into the waveform. It has been degrading audio in every mode since the bridge
was written - speech just masks it.

The two fixes are different halves and neither subsumes the other:

  yours   decay lastSample toward silence on underrun, instead of hard zeros.
          Removes the step discontinuity, so the gap stops clicking.
  mine    prefill a few frames and re-prime after a dry spell.
          Stops the gap repeating once per render quantum until frames return.

I have kept both, with your decay credited to you, so nothing is lost whichever
way the merge goes.

The reason for the warning: on feature/cw-reader the shared browser audio
modules moved out of the app and into the shared core, so this file now lives at
core/js/audio/audio-playback.js. wwwroot/js/audio/audio-playback.js is a
CopySharedCoreJs build artefact from that point on, and it is gitignored.

That means a merge of this PR after the move shows modify/delete on that path.
Resolving it as a delete is the tempting reading - the path is generated now -
and it would drop your fix with no conflict marker, no build error and nothing
in the diff to notice. That is why I folded it into core ahead of time rather
than leaving it to the merge.

Nothing needed from you unless you would rather do it yourself. The
audio-playback.js hunk in this PR is now redundant; dropping it would make the
merge trivial, but leaving it is fine too since core already carries the change.

One other overlap, not dangerous but worth knowing. This PR and
feature/cw-reader both touch Services/Audio/RadioAudioBridgeService.cs. Your
device grace window, EnsureCodec and session-handoff reuse are in different
regions from the RxFrameCaptured hook the CW decoder taps, so they should merge
cleanly, but it is worth a build after whichever lands second.

fabiojvalente:
Thanks for the thorough review, Colin — addressed in 3147bb6:

1. **Grace-close race** — token re-check and `CloseDevicesAndCodecOnAudioThread` now run inside `AudioDeviceEnumerator.Invoke`, and device reuse vs reopen is decided in the same queue so a late queued close cannot dispose streams under a reconnecting session.
2. **Playback callback** — `EnsureCallbackBuffer` copies on grow so resampled samples are not lost when `_cbPlayMono` resizes.
3. **Stray docs** — removed `required_before_merge.md` and `video_debugging.md`.
4. **`HasActiveSession`** — now tracks `_connectionId` (busy until `Release`), matching `TryAcquire`.
5. **`ListDevices`** — single code path with `cachedWhileStreaming` flag.
6. **PR description** — updated to drop the #122 CAT-polling claim and call out the jitter (3→6 frames) and levels (100→200 ms) trades.

Build clean, 30/30 tests on `net10.0`. Ready for your bench run on the '101.

mm5agm (approved):
Approving. Bench run done on the FTdX101 (RX only), plus a clean build and
test run here.

Build: clean, both TFMs (net10.0 and net10.0-windows), 0 warnings.
Tests: 30/30 on net10.0, confirmed independently.

Bench results
-------------

| Test                                    | Result |
|-----------------------------------------|--------|
| Pop out audio, then navigate main window | Pass   |
| Rapid disconnect / reconnect             | Pass   |
| Enumerate devices while streaming        | Pass - no native crash |
| Session ownership across browser tabs    | Pass - correctly refused, self-clears |
| Added latency by ear                     | Pass - no audible penalty |

The device enumeration one is the headline: no host crash when Settings is
loaded while a WASAPI capture stream is running. Routing the PortAudio
control plane onto a single dedicated thread and serving the cache while
streams are open is the right shape for this, and the comment explaining
why is worth having in the file.

Locking the mic and codec selects while streaming is also an improvement
on its own terms. Those controls previously accepted a change mid-stream
and silently did nothing except update localStorage, so greying them with
"Stop and reconnect to apply" replaces a control that lied with one that
tells the truth.

Three things surfaced during the bench. None of them are caused by this PR
and none of them block it:

1. Audio level drops when a signal comes up out of the noise. This is the
   radio's own AGC - the front panel speaker does exactly the same thing.
   Not a software issue at all.

2. Pressing the pop-out button when a pop-out window already exists gives
   no feedback. window.open reuses the named window "ywc-remote-audio" and
   refocuses it, which looks like nothing happening if the window is behind
   the main one or on another monitor. Pre-existing, cosmetic, worth a
   one-line status message some time. Note this PR makes the surrounding
   code better, not worse - dropping the 150 ms settle shortens the gap
   between the stop and the window.open.

3. Unrelated to audio: the CW decoder tap sits downstream of the RX gain
   and its clamp, so the decoder sees samples already scaled and hard
   clipped at +/-1.0. That is mine to chase on the CW branch.

Merging into develop.

PR #113, opened 2026-08-18
CW reader plan: decoder in Core, and lifting audio capture out of the Remote Audio session

mm5agm:
Docs only. No code, no version bump. This is the agreed plan for a CW reader
in YWC and IWC, opened as a PR so the design can be argued with before any of
it gets written.

Full plan: `docs/design/cw-reader-plan.md`

## The bit that needs a second opinion: the Remote Audio capture

This is the one part of the plan that reaches into someone else's work, so it
is the reason this is a PR rather than a commit.

Today the PortAudio input stream is owned by the Remote Audio session.
`RadioAudioBridgeService.OpenDevices` runs when a browser opens the audio
WebSocket, and `AudioSessionManager` rejects a second connect as busy. The
capture device therefore exists only while somebody is actually listening
remotely.

That is exactly right for what the bridge was built to do. It does not work for
a decoder. A CW reader that only runs while a Remote Audio session happens to
be up is not the feature, and it would not run at all headless, which is where
the Pi and Docker users are.

So Phase 2 proposes lifting the capture side out of the bridge into a
`RadioAudioCaptureService`:

- reference-counted start/stop, so the device opens when the first consumer
  asks for it and closes when the last one leaves
- a multicast frame event, fired at the 480-sample accumulator boundary that
  already exists in `RadioAudioBridgeService` (around line 543). That accumulator
  is already doing the framing, so it is the natural seam rather than a new one
- the bridge becomes one subscriber among others

The TX and playback path is not touched. Neither is the Opus encoding, the wire
protocol, or the session model for remote listening. From a remote listener's
point of view nothing changes: the same device opens, the same frames arrive.
What changes is that the reader can also subscribe, and either consumer can run
without the other.

Fabio, this is your code (#92 and the audio work on develop), so if you would
rather this were done a different way, or you have plans for that area that this
would collide with, this PR is the place to say so. I would rather find that out
now than after the refactor is written. If the extraction looks wrong to you, an
alternative is to keep the bridge as owner and give it a subscriber hook, but
that keeps the "decode only works while someone is listening" limitation, which
is what I am trying to get away from.

## Three other findings that set the design

**Neither radio will give us decoded text.** The FTdx101 and the IC-7300 MkII
both decode CW themselves and both expose the settings over the wire (FTdx101
menu `01 CW DECODE BW`, Icom `02 49` to `02 51`), but neither has a command that
reads the characters back. Icom's USB (B) port can be switched to emit decode
text via `1A 05 00 94`, but the setting is `00=RTTY Decode, 01=CI-V`, so RTTY
only. There is no passthrough shortcut on either brand. Worth having checked,
because a passthrough would have been a fraction of this work.

**IWC's ZIN cannot mirror YWC's.** YWC's is one line, `ZI{P1};`, and the radio
nudges its own VFO. The IC-7300 MkII has the function, CW Auto Tuning with a
front-panel AUTOTUNE key, but CI-V only exposes `1A 05 00 58`, which is the key
assignment and not a trigger. So IWC has to compute it: measure the tone,
compare against the configured CW pitch, shift the VFO. That is a bonus rather
than a cost, because the tone estimator the decoder needs is the same one
zero-in needs. One engine serves both.

**IWC has no `Services/Audio/` at all.** It needs a capture-only path built,
with no Opus and no WebSocket, so a small fraction of what YWC already has.

## Where the code goes

Most of it lands in `Radio_Web_Control_Core`, because a Morse decoder does not
know what a radio is. Core gets the decoder engine, the Morse table, the pure
zero-in maths, the transcript writer and an ADIF record writer, all behind an
`ICwAudioSource` seam. Core stays on `net10.0` with no package references, so
YWC's CAT-only macOS/Linux target keeps building. Opening an audio device and
sending anything over CAT or CI-V stays in the applications.

## Phases

| | | verified by |
|---|---|---|
| 0 | branches plus this document | done |
| 1 | Core: decoder engine, zero-in, Morse table, synthetic-audio test suite. No app wiring. | `dotnet test` gives an accuracy number against WPM and SNR |
| 2 | YWC: capture extraction, reader service, SignalR, pop-out page | real CW off the FTdx101MP |
| 3 | YWC: Reader Mode, transcript, ADIF QSO save | as above |
| 4 | Core to IWC: subtree push, IWC capture service, reader UI, software ZIN | real CW off the IC-7300 MkII |
| 5 | USER_MANUAL section, README notes | |

Phase 1 exists to decide whether this is worth shipping at all. CW decoders are
usually mediocre and it is hard to judge how mediocre by ear, so that phase
produces a measured accuracy figure against synthetic Morse at set WPM and SNR
before any UI is built. If the numbers are poor at realistic SNR, that is known
before there is anything to throw away.

One item cannot be settled from the manuals and has to be bench-verified: the
zero-in sign convention, CW-U against CW-L. Getting it backwards gives a
zero-in that runs away instead of converging, so it ships with a deadband and a
clamp.


mm5agm:
@fabiojvalente flagging this one for you, since the only part of it that
touches your work is the audio capture.

Short version: the CW decoder needs PCM whether or not anyone is listening
remotely, and right now the capture device only exists while a Remote Audio
session is open. The plan proposes lifting the input side out of
`RadioAudioBridgeService` into a reference-counted service with a multicast
frame event, hanging off the 480-sample accumulator that is already there,
with the bridge becoming one subscriber. TX, playback, Opus and the wire
protocol are untouched, and a remote listener should not be able to tell the
difference.

No code written yet, and no rush on a reply. If you would rather that area
were done differently, or you have something planned for it that this would
collide with, now is the cheap moment to say so. The alternative I considered
and rejected is in the PR body.

PR #118, opened 2026-08-18
Stop the Radio Display showing a tiled picture on merging grabbers

mm5agm:
## The fault

My USB HDMI grabber renders every downscaled MJPEG mode it advertises as a
tiled repeat of the radio's screen. Only its largest mode is correct:

| Mode requested | Result |
|---|---|
| 800x600 (4:3) | 5 copies |
| 1280x960 (4:3) | 2 copies |
| 1280x720 (16:9) | 2 copies |
| 1920x1080 (16:9) | clean |

It is not an aspect-ratio problem - 1280x720 is 16:9 and still fails. It is
that every mode below the device's native one runs through its internal
scaler, and that scaler is broken. The Windows Camera app looked fine
throughout because it takes 1920x1080; our pin ranking prefers 4:3.

## Why nothing downstream could fix it

Good and bad samples are both perfectly well-formed JPEGs:

| | Clean | Tiled |
|---|---|---|
| SOF | 1280x960 4:2:2 | 1280x960 4:2:2 |
| SOS / EOI | 1 / 1 | 1 / 1 |
| Restart markers | 0 | 0 |
| Scan length | 177,465 bytes | 257,392 bytes |

One image, one scan, ending exactly at its EOI. The extra pictures are inside
that single entropy-coded scan, so there is nothing to split or trim.

What does mark a bad sample is its tail: it carries the start of the next
frame past its own EOI. Over comparable runs, the failing mode produced 9
rate-limited warnings (one logged per hundred, so ~900 samples) and the
working mode produced zero. That makes three such samples conclusive rather
than borderline.

## The change

1. Drop merged samples instead of publishing them. A tail beginning with
   another SOI means the driver put several pictures in one sample. A tail
   without an SOI is only padding, so it is trimmed and published as normal.
2. After three merged samples, reopen on the device's largest MJPEG mode and
   scale down in software. Nothing is hardcoded to 1920x1080 or to this
   grabber - the fault is detected by signature, and the native mode is
   whatever that device reports.
3. Drop 4-byte `FF D8 FF D9` samples - valid JPEGs containing no image. This
   grabber emits them continuously in every mode including the good one,
   nearly 3000 in a single run. Publishing them overwrote real frames, which
   is how the panel could sit black while the badge read a healthy green
   14.7 fps. Dropping them also lets the existing stall watchdog report an
   honest disconnect.
4. Trim by walking the marker structure instead of scanning backwards for
   `FF D9`. That byte pair occurs freely inside marker payloads - an APPn
   thumbnail is itself a JPEG ending in one - so the backwards scan could
   stop short or run long.

`YWC_VIDEO_FORCE_PIN=WxH` pins capture to a single MJPEG mode. That is how
the four modes in the first table were compared.

## Result on the bench

```
20:48:14.011  sample held more than one picture - Dropped   <- occurrence 1
20:48:14.722  device is merging frames at 1280x960
20:48:16.562  negotiated 1920x1080 passthrough+scale
20:48:16.575  first JPEG published (1280x720)               <- first frame shown
```

No "first JPEG published" line for 1280x960: not one corrupt frame reached
the panel. The display stays blank for about two seconds, then the picture
arrives correct and stays correct - frames hold a steady 144-154 KB with zero
merge warnings after the switch.

## Scope

Linux and Media Foundation sessions take the interface default of "not a
merger" and are behaviourally unchanged. If those dongles do the same thing,
the detection lifts into the V4L2 session as it stands.

## Test plan

- [x] FTDX101MP + USB HDMI grabber on Windows: starts on 1280x960, detects,
      reopens on 1920x1080, no corrupt frame published
- [x] Sustained clean capture verified by frame inspection, not just status
- [x] `dotnet build -f net10.0-windows` clean, 13 tests pass
- [ ] Linux Radio Display unaffected (no code path changed, worth a look)

Bench-tested on my own rig only - one grabber, one radio. The detection
threshold is deliberately conservative, but a second pair of eyes on whether
any healthy device could legitimately pack two pictures into one sample
would be welcome.

### DIFF
--- a/Services/Video/IJpegCaptureSession.cs
+++ b/Services/Video/IJpegCaptureSession.cs
@@ -1,4 +1,4 @@
-namespace Yaesu_Web_Control.Services.Video
+﻿namespace Yaesu_Web_Control.Services.Video
 {
     /// <summary>
     /// Compressed-MJPEG capture that copies JPEG bytes as-is (DirectShow pin
@@ -12,5 +12,14 @@ namespace Yaesu_Web_Control.Services.Video
         double DeviceFps { get; }
         bool TryReadJpeg(out byte[]? jpeg);
         bool TrySetFrameRate(int targetFps);
+
+        /// <summary>
+        /// True once the device has been seen packing several pictures into one
+        /// sample. Some HDMI grabbers do this in every mode that runs through
+        /// their internal scaler: the JPEG is well formed but its scan holds more
+        /// than one frame, which renders as a tiled, repeated picture. Capture
+        /// switches to the device's native mode when this is reported.
+        /// </summary>
+        bool DeviceMergesFrames => false;
     }
 }
--- a/Services/Video/VideoCapturePinRank.cs
+++ b/Services/Video/VideoCapturePinRank.cs
@@ -1,4 +1,4 @@
-namespace Yaesu_Web_Control.Services.Video
+﻿namespace Yaesu_Web_Control.Services.Video
 {
     /// <summary>
     /// Radio Display pin ranking shared by Windows DirectShow MJPEG, Media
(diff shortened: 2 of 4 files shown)


mm5agm:
Independent confirmation from a second bench run today, by accident rather than design.

I had a build off develop (without this fix) running for unrelated work. The report against it was "2 displays big green blob instead of a spectrum" - the original fault exactly. Rebuilding on the same machine, same grabber, same session from this branch gave "perfect".

That is a clean A/B on identical hardware minutes apart: develop tiles, this branch does not.

It does not widen the evidence to a second grabber, which remains the gap in the test plan above.

PR #119, opened 2026-08-18
Keep both VFO panels live on single-receiver radios

fabiojvalente:
## Summary
- On single-receiver radios (FTdx10, FT-710, FTDX3000) both VFO panels stay fully editable — the inactive panel is no longer greyed out or blocked.
- RX/TX selectors follow the radio’s front-panel colours (RX green, TX red) and keep TX coherent when RX moves: non-split, TX follows RX; already split, TX stays put until RX lands on it.
- FTDX3000 uses FR/FT (including FT2 when split clears); FTdx10 / FT-710 / FT-991A also send ST to engage or clear split.
- Save to Mem and QMB sit under the filter/audio scope, centred in the gap between the band buttons and the AGC column.
- **Mode on the inactive VFO no longer rewrites the active one.** `MD` is per-VFO even on single-receiver radios (P1 0=MAIN/VFO-A, 1=SUB/VFO-B). It was being sent as `MD0` for both panels via the P1=0-Fixed path, so changing mode on VFO B while A was active also set A.

### Screenshots
image

## Test plan
- [x] **Single-receiver (FTdx10 / FT-710 / FTDX3000):** both VFO panels stay live and editable; changing the inactive VFO does not grey it out.
- [x] **RX/TX selectors:** RX is green, TX is red. With split off, changing RX also moves TX. With split on, TX stays until RX is set to the same VFO (split clears).
- [ ] **FTDX3000:** RX B / both-on-B does not show reverse split (RX B / TX A). Split still works from the TX selector.
- [x] **FTdx10 / FT-710:** ST engages/clears split with the TX selector.
- [x] **Layout:** Save to Mem (and QMB on VFO A) sit under the filter/audio scope, centred between the band buttons and AGC. Dual-receiver (FTdx101) layout still looks right.
- [ ] **Mode (FTdx10):** with VFO A active, set VFO B to a different mode — A stays put; swapping to B shows the mode you set. Same the other way around.
- [x] Screenshots — to be added.

### DIFF
--- a/Controllers/CatController.cs
+++ b/Controllers/CatController.cs
@@ -976,57 +976,72 @@ namespace Yaesu_Web_Control.Controllers
                 if (recv != "A" && recv != "B")
                     return BadRequest(new { error = "Invalid receiver specified" });
 
-                await _catClient.SendCommandAsync($"MD{VfoP1Outgoing(recv)}{request.Mode};", "User");
-                if (VfoIsB(recv)) _radioStateService.ModeB = displayMode;
-                else               _radioStateService.ModeA = displayMode;
+                // MD is per-VFO even on single-receiver radios (FTdx10 CAT
+                // manual: P1 0=MAIN/VFO-A, 1=SUB/VFO-B). VfoP1Outgoing would
+                // send MD0 for both panels and change the active VFO's mode
+                // when the operator edited the inactive one.
+                bool requestedB = recv == "B";
+                string mdP1 = RadioCapabilities.ModeP1(recv);
+                await _catClient.SendCommandAsync($"MD{mdP1}{request.Mode};", "User");
+                if (requestedB) _radioStateService.ModeB = displayMode;
+                else            _radioStateService.ModeA = displayMode;
 
-                // Re-apply Contour and APF state — mode changes on the FTdx101 cause the
-                // radio to restore its per-mode Contour/APF settings, overriding what we have set.
-                var modeSettings = await _settingsService.GetSettingsAsync();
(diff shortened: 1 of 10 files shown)


fabiojvalente:
@mm5agm would you be able to test this on a dual receiver unit and ensure behaviour isnt affected on those ?

Passed the tests on the ftdx10 and includes a bugfix on when changing the mode on the standby vfo

mm5agm:
@fabiojvalente Tested on the FTdx101MP - dual receiver, so the case you could
not cover yourself.

Mode stays independent in both directions:

- MAIN active, changed SUB's mode: SUB moved, MAIN did not.
- SUB active, changed MAIN's mode: MAIN moved, SUB did not.
- The radio's own front panel agreed with the app after each.

Split, and the SUB receiver switched off and back on again, both behaved as
they did before. No meter stutter with the meters running.

Worth saying where I aimed the test, in case it changes what you want covered.
On a dual-receiver radio the new `ModeP1` returns exactly what `VfoP1` already
returned - 0 for A, 1 for B - so the routing change is a no-op here by
construction, and "nothing moved" is the expected result rather than a
surprising one. The one thing that is genuinely new for this radio is `MD1;`
joining `MD0;` in the poll set: an extra command every cycle on a port already
carrying the 10 Hz meter poll. That is what I was watching the meters for, and
it is clean.

The one thing I have not done is a long soak, so if the extra poll command
costs anything it would be over hours rather than minutes. Everything above
was minutes, not seconds.

73 Colin MM5AGM

PR #122, opened 2026-08-19
Push meter updates over SignalR and reduce CAT poll / persistence overhead

fabiojvalente:
### Biggest win: remove browser-side status polling

Every open browser tab was running `setInterval(fetchRadioStatus, 500)` — hitting `/api/cat/status` twice a second and repainting the whole UI from the response, duplicating what SignalR already pushes. **Deleting that poll loop is the single largest change in this PR** and a bigger saving than the CAT-side work.

First paint is still covered: `GetClientStateSnapshot()` sends Band / Mode / Antenna / RoofingFilter / Power on connect, and the existing SignalR handlers perform the same DOM updates the poll used to drive.

### Reduced flash wear (Raspberry Pi / Docker)

Relevant for Pi setups running YWC in Docker with config on the SD card (`XDG_CONFIG_HOME=/data` → `MM5AGM/Yaesu Web Control/`).

**Before:** every `SetField()` after init called `Save()` synchronously — including high-frequency meter updates (`SMeterA`, `SMeterB`, `PowerMeter`, `CompressionMeter`, `IsTransmitting`, etc.). Meter polling could trigger a **full rewrite of `radio_state.json` several times per second**, even though those meter values are not stored in that file. `GetSettingsAsync()` also re-read `appsettings.user.json` from disk on every meter poll cycle (~2 Hz).

**After:**
- Only properties that belong in `RadioState` mark the file dirty; meter/transient fields no longer trigger persistence.
- Writes are **debounced** to at most once every **5 s** via a background flush (atomic `.tmp` + rename). `radio_state.json` stays **indented** for hand debugging.
- Settings are **cache-first** with `FileSystemWatcher` invalidation — one read at startup instead of ~2/s. On Docker/Pi bind mounts the watcher may not fire for host-side hand-edits; prefer the Settings page, or restart after editing the file by hand (noted in USER_MANUAL).

On a Pi running 24/7, this turns constant meter-driven rewrites into occasional batched writes when real operator settings change.

### RF power scale by radio model (`6e38960`)

On 100 W radios (e.g. FTdx10) the Power slider and Power meter dial were both drawn **0–200 W**, so full output sat in the middle of the arc. Two causes:

1. **Slider regression on this branch** — retiring `/api/cat/status` dropped the poll that seeded `state.radioModel`; the deferred `window._radioModel` assignment ran after the slider was sized, so it always fell back to 200 W.
2. **Dial never right** — `PowerGauge` had been hard-coded `maxValue: 200` since the FTdx101MP-only days (same on develop).

`RadioCapabilities.MaxPowerWatts` is now the single source for slider max, gauge dial, meter clamp (`FTdx101Meters`), and `CatController.SetPower` validation. Index server-renders `data-max-power` so first paint does not depend on JS timing. Also fixes FTDX5000: docs/slider said 200 W while SetPower rejected anything above 100 W.

Introduced in commit [`6e38960`](https://github.com/mm5agm/Yaesu_Web_Control/commit/6e38960d60aa0cd784fe163b1c0b5ccb1b5fa091) (*Scale RF power to the radio's own rating, not the FTdx101MP's*). That commit also added `Dispose()` on the test `FakeSettingsService` so the test project builds again after `ISettingsService : IDisposable`.

### Other changes

- **SignalR push for meters** — `RadioStateService.GetClientStateSnapshot()` now includes S-meter and TX meter values; the frontend handles `SMeterA`/`SMeterB` and meter properties via existing SignalR handlers instead of polling. S-meter handlers are try/catch-guarded so an early update cannot abort the rest of the message handler.
- **Retire client-side status polling** — Removes `fetchRadioStatus()`, the `setInterval`, `pausePolling`/`resumePolling`, and the unused `updateModeAndAntennaButtons` leftover.
- **Tiered meter polling** — `MeterPollingService` restructured into fast (TX + S-meters every cycle), TX-only (power/ALC/compression/SWR while keyed), and slow (VDD, temperature, antenna, frequency backstop) tiers. On a single-receiver radio receiving, CAT traffic drops from ~20 cmd/s to ~14 cmd/s by gating TX-only meters — directly relevant to shared-CAT / VSPE setups (#78).
- **`MeterPollIntervalMs`** — Settings control for the **minimum cycle period** (delay between cycle starts; default 200 ms, clamped 50–1000). Each loop subtracts cycle elapsed time from the interval before delaying, so the configured value is a real period rather than “wait after work”.
- **Wall-clock thresholds** — Antenna poll, frequency backstop, disconnect detection, S-meter zero-hold (~1 s, matching the old 3×500 ms behaviour), and meter-borrow settle windows converted from cycle counts to wall-clock intervals.
- **Debounced persistence** — `RadioStatePersistenceService` converted to a `BackgroundService` with 5 s periodic flush and atomic writes; `SetField` only marks dirty for properties that persist to `radio_state.json`.
- **Settings cache** — `SettingsService` cache-first read with `FileSystemWatcher` invalidation on external `appsettings.user.json` changes.
- **Radio wins on connect (behaviour change)** — Stopped restoring persisted antenna (`AN0`/`AN1`) to the radio at startup, consistent with Mode (#38) / RF Power (#35) / MIC Gain (#16). The radio’s current selection is read and followed instead.
- **TX meter zeroing** — Frontend immediately zeroes power/SWR/compression/ALC/IDD on TX-off; backend zeroes all TX-only meters atomically on the `IsTransmitting` false transition to prevent stale peak readings during the TX debounce window.

### Review follow-ups (Colin MM5AGM)

Addressed on this branch after the initial review:

- Test build: `FakeSettingsService` implements `IDisposable` (in `6e38960`)
- Meter interval subtracts elapsed work; Settings / USER_MANUAL / CLAUDE reworded as cycle period
- S-meter zero-hold restored to ~1 s (was 2.3 s)
- FTdx101Meters smoothing comment corrected for ~4–5 Hz TX cadence
- CLAUDE.md CAT codes corrected (`MS13`+`RM0` vs `RM3`/`RM6`; `RM8` is VDD)
- `WriteIndented` restored on `radio_state.json`
- Unguarded `updateSMeter` wrapped; dead `updateModeAndAntennaButtons` removed
- Docker bind-mount FileSystemWatcher caveat documented

## Files changed

| Area | Files |
|------|-------|
| Backend polling & state | `MeterPollingService.cs`, `RadioStateService.cs`, `RadioStatePersistenceService.cs`, `RadioInitializationService.cs`, `Program.cs`, `RadioCapabilities.cs`, `CatController.cs` |
| Settings | `ApplicationSettings.cs`, `SettingsService.cs`, `ISettingsService.cs`, `Pages/Settings.cshtml` |
| Frontend | `wwwroot/js/ui/site.js`, `FTdx101Meters.js`, `gauge.js`, `radio-scope.js`, `Pages/Index.cshtml` |
| Tests | `RadioMaxPowerTests.cs`, `VideoCaptureServiceHaltTests.cs` |
| Docs | `CLAUDE.md`, `USER_MANUAL.md`, stale comment fixes in `ScopeController.cs` |

## Test plan

- [x] Build both TFMs: `dotnet build Yaesu_Web_Control.csproj`
- [x] Run unit tests: `dotnet test Tests/YaesuWebControl.Tests/YaesuWebControl.Tests.csproj` (30/30)
- [ ] Open `http://localhost:8080` — confirm meters update live without `/api/cat/status` polling (Network tab)
- [ ] Tune frequency from web UI, front panel, WSJT-X, and a second browser tab — all stay in sync
- [ ] Key TX — power, SWR, compression, ALC, IDD respond; unkey — all TX meters drop to zero immediately (no stuck compression/ALC peak)
- [ ] Change antenna on front panel — web UI follows within ~2 s; restart with a different front-panel antenna and confirm YWC does **not** overwrite it
- [ ] Brief S-meter zero flash recovers in ~1 s (not ~2.3 s)
- [ ] Debug log: cycle `elapsed + delay ≈ MeterPollIntervalMs`
- [ ] Adjust **Meter Poll Interval (ms)** on Settings — verify cadence; `radio_state.json` remains indented after a real settings change
- [ ] Reconnect SignalR (refresh page) — meter snapshot replays correctly on load
- [ ] Power slider / gauge / POST `/api/cat/power` match model rating (100 W FTdx10 / 200 W FTdx101MP or FTDX5000) — see `6e38960`

## Notes

- Default `MeterPollIntervalMs` is 200 ms (minimum cycle period) — conservative for shared-CAT setups; raise it if digital modes collide on the bus.
- Antenna “radio wins on connect” is a deliberate behaviour change for release notes.
- No version bump in this PR.


fabiojvalente:
https://github.com/mm5agm/Yaesu_Web_Control/commit/2fb2fe24140fe139bfb0842fa8fb540e9fed3692#commitcomment-196866573

mm5agm:
Fabio - I read `rnd/` as a branch you were proposing and reviewed it as though it were up for merge. It was a scratch branch you had pushed so the bench could build it, and my comment did not treat it that way. Sorry for that.

Draft is exactly the right state for it, and there is no hurry from me. Treat everything in that commit comment as notes for whenever you pick it back up, not a list of things blocking you - the only one I would not leave sitting is the test build, because it will bite anyone else who touches the repo in the meantime.

The PR description is excellent, by the way. Leading with the `/api/cat/status` poll removal is right - that is the headline change and I had missed it for months.

73, Colin MM5AGM

fabiojvalente:
@mm5agm give it a run when you can :)

mm5agm:
Hi Fabio,

Following up on the review comment now that this is a PR. No rush on any of it -
draft is the right state and nothing here is blocking.

First, thank you for this one. The two structural changes are exactly what I had
on my own list: the persistence service becoming a BackgroundService with a dirty
flag and an atomic write solves the SD-card wear on the Pi rig, and moving the
meter poll from cycle-counting to wall-clock deadlines fixes the drift I have been
living with. The win you did not mention in any commit message is arguably the
biggest: deleting setInterval(fetchRadioStatus, 500) from site.js. Every open
browser tab was polling /api/cat/status at 2 Hz alongside SignalR, and first paint
is still covered by GetClientStateSnapshot(). Good catch.

I also checked the two things most likely to bite and both are handled correctly:
the hosted registration uses the factory form so there is one instance rather than
a hosted one plus a separate singleton, and Directory.CreateDirectory runs before
the FileSystemWatcher, so a first-run install with no AppData folder will not throw
at DI construction.

ONE BLOCKER

The branch breaks the test project:

  dotnet test Tests/YaesuWebControl.Tests

CS0535 - FakeSettingsService does not implement Dispose(), from
ISettingsService : IDisposable. Still failing on head 3919447 as of today.

Worth flagging why it slipped: both app TFMs build clean, because the test csproj
is not referenced by the app csproj. Same class of slip as the CS0104 on #96. I
have started building the test project separately when reviewing a branch.

ON THE METER POLL RATE - a correction that is mine, not yours

You were right that CLAUDE.md was wrong, but I do not think 1.3 Hz is the develop
figure either. MeterPollingService on develop is Task.Delay(500), so 2 Hz, and the
code comments agree ("every meter-cycle (2 Hz)"). Your 1.3 Hz looks like it
describes the tiered behaviour on this branch rather than what is on develop today.

Since you already have that section of CLAUDE.md open, could you land the correct
wording as part of this PR? I would rather you fixed it once here than have me
edit the same lines on develop and hand you a conflict for no reason. CLAUDE.md
being wrong about a core figure costs us both, so I am glad you spotted it.

THREE THINGS I WOULD LIKE TO SETTLE BEFORE IT MERGES

1. The default poll interval going 500 ms to 200 ms.

   My preference is to keep 500 ms as the shipped default and let people opt in to
   200 ms through the new MeterPollIntervalMs setting. The mechanism is good and I
   want it - it is only the default I am cautious about. 200 ms means 2.5x the CAT
   traffic for every user by default, and the people it bites hardest are the ones
   sharing the port: Giovanni's VSPE/FTDX3000 setup in #78, and anyone running
   WSJT-X or rigctld alongside. With 100-odd users I would rather the safe value
   was the one that ships and the snappy S-meter was a deliberate choice.

   Happy to be argued out of this if you have measurements from your bench that
   say the extra traffic is harmless on a shared port - you have tested more
   configurations than I have.

2. WriteIndented = true was dropped from the radio_state.json serializer.

   Small thing, but it makes the file one long line, and I read that file out of
   people's bug reports fairly often. Could we keep it indented? If the write cost
   is what motivated dropping it, I would rather pay it.

3. CLAUDE.md edits.

   No objection at all - the content was right and I would rather you fixed a wrong
   file than left it. Only asking that changes to that one get called out in the PR
   description, since it is the file that steers how Claude behaves for both of us
   and a silent change there is easy to miss.

MEASUREMENTS I TOOK, IN CASE THEY ARE USEFUL

  MeterPollIntervalMs is a delay between cycles rather than a rate - the Stopwatch
  is logged but never subtracted, so the real period is work + interval.

  CAT traffic after the change:
    FTdx101 RX          about 17 cmd/s, roughly unchanged (S-meter 2.5x faster)
    FTdx101 TX          about 18 to about 37 cmd/s
    single-receiver RX  about 20 to about 14 cmd/s, which helps #78

One process note while I am here, and it is a question rather than a complaint:
would you rather I reviewed rnd/ branches or left them alone? I reviewed this one
as if it were up for merge before realising rnd/ meant research and you had only
pushed it so your remote bench could build it. Happy either way - just tell me
which you prefer and I will stick to it.

73 Colin

mm5agm:
Gave it a run on the FTdx101MP this morning. Short version: it builds clean,
the tests pass, and nothing I saw at the rig was down to this branch.

## Build and test, on e063e5f

```
dotnet test                30 passed, 0 failed   (the CS0535 blocker is gone)
dotnet build -f net10.0-windows    0 warnings, 0 errors
dotnet build -f net10.0            0 warnings, 0 errors
```

Confirmed on the running app rather than just in the source:

- `WriteIndented` really is back. `radio_state.json` on disk is pretty-printed again.
- Connected to COM4 at 38400 8-N-2, listening on 8080, both SDR workers up.
- Zero `[ERR]` or `[FTL]` in 22 minutes of live running with a lot of TX.
- No `MeterPollIntervalMs` in my user settings, so this ran on the branch
  default of 200 ms. What I was looking at is the thing under review.
- The CLAUDE.md meter-rate wording now matches what the code actually does.

## At the rig

RX and TX transitions, band changes and general operating all looked normal.

The S-meters greying and dropping to zero for the duration of an over is
correct and unchanged by this branch, for anyone reading along: the radio
latches SM0 and SM1 at key-down, so those are not live readings, and greying
them is more honest than drawing a frozen needle.

I also had my two receivers on slightly different frequencies while comparing
their S-meters, which accounts for a difference I thought I was seeing. Both
VFOs share one calibration table, so YWC is not treating them differently.

## One fault found, and it is not yours

My antenna is faulty at the moment, so my SWR pegs at full scale. With that
happening, the SWR meter freezes: it stays at 0 for the whole transmission and
then shows a stale value on receive.

I chased it down and it is a pre-existing bug on `develop`, from 601b111, in the
`SWRMeter` spike filter in `RadioStateService.cs`. Once the value hits 255 the
filter rejects every real reading that follows, so the value latches, no SignalR
update is sent, and the gauge never redraws. My log has five rejected readings
that never reached the display.

Nothing to do with this PR, and it should not hold it up. I have filed it separately as #124. Because `RadioStateService.cs` is a file this branch also edits, I
will fix it **after #122 merges** rather than handing you a conflict.

## On the power scaling in 6e38960

Good catch, and the right shape. Worth flagging that my FTdx101MP is rated
200 W, so my rig is a null test for it - the change only bites on the 100 W
models. It is covered by tests, but real confirmation wants an FTdx10 or FT-710
owner, which I will line up alongside the next thing that needs bench testing.

## My one remaining review point, withdrawn

I asked earlier for the default `MeterPollIntervalMs` to stay at 500 ms. Having
now run it at 200 ms on the radio, I am withdrawing that. The S-meter is not
twitchy, the RX and TX transitions are clean, and 22 minutes of use with plenty
of transmitting produced no CAT trouble and nothing in the log. Keep 200 ms.

That was the last thing outstanding from my review, so from where I sit this is
good to merge.

fabiojvalente:
Good morning @mm5agm
Power scaling for 100w was tested and validated on the dx10 while creating this branch ✅ 

Are there any blockers to get this one merged ?


I've been having a local branch with this PR and all others merged running for hours, no hicups or problems so far :)

mm5agm:
Morning Fabio.

That answers the one thing I could not check myself. My '101 is a 200 W radio,
so `MaxPowerWatts` is a null test on my bench - it was always going to need a
100 W radio to prove. Validated on the FtdX10 is exactly what I wanted to hear,
and it comes off my list.

**No blockers from my side.** All three of my review points are closed:

1. **Default poll interval.** I asked for 500 ms and I am withdrawing that. I
   ran it at the 200 ms default on the radio yesterday - S-meter is not twitchy,
   RX and TX transitions are clean. Keep 200 ms.
2. **`WriteIndented`.** Restored, and I confirmed it on disk rather than just in
   the source - `radio_state.json` is pretty-printed again.
3. **CLAUDE.md meter rate.** Fixed, and the wording now matches what the code
   actually does.

Alongside that: 30/30 tests passing on `e063e5f`, both TFMs building 0 warnings
0 errors, and 22 minutes live on the rig with nothing in the log.

The SWR fault I filed as #124 is **not** a blocker either. It is pre-existing on
`develop`, from `601b111`, and nothing to do with this branch - I only tripped
over it because my antenna is faulty at the moment and my SWR pegs at full
scale. I am deliberately holding that fix until this merges so you do not get a
conflict in `RadioStateService.cs`.

So the only thing actually stopping it is mechanical: **the PR is still marked
as a draft**, and GitHub will not merge a draft. Click "Ready for review" and it
can go in.

One note on your local branch, not a concern, just so we are not both assuming
the same thing: you have been running this PR *plus all the others* merged
together, and I ran this branch on its own against the radio. Between the two we
have both the combined state and this change in isolation covered, which is
better than either on its own.

PR #138, opened 2026-09-12
Radio Display: prototype click-to-tune and clickable controls on the captured TFT

mm5agm:
## What this is

A prototype for using the Radio Display (captured TFT) as a control surface rather than a picture - the first step toward using it instead of an SDR panadapter.

@fabiojvalente - this is built on your Remote Video work, so I'd like your eye on it before it goes anywhere. It's deliberately a prototype: one layout, measured from a screenshot, with the tools to correct it on the bench rather than a claim that the numbers are right.

## What it does

- **Spectrum / waterfall**: hover shows the frequency under the cursor; click tunes `FA`/`FB` to it.
- **ANT / ATT / IPO / R.FIL / AGC** readouts: click cycles the setting over the existing `/api/cat` endpoints.
- **CURSOR / SPAN / 3DSS / HOLD** soft-buttons: drive the CAT scope control (`SS`).
- **MONO / MULTI / EXPAND / MEM CH**: hover text says there is no CAT command for them, click does nothing.

## How the frequency is worked out

The radio draws the VFO marker at the VFO frequency in CENTER, CURSOR and FIX placement alike, so

```
f(x) = VFO + (x - markerX) * span / plotWidth
```

needs no knowledge of where the axis starts. The marker is found by drawing the current MJPEG frame to an off-screen canvas and scanning the spectrum rows for the tallest run of saturated red. `span` comes from the live `SS` state the scope panel already holds; `VFO` from `FrequencyA`/`FrequencyB` for the band the scope is showing.

## Fit / Fill / docking

The overlay is positioned to match the *drawn* image inside the `` box (object-fit contain/cover maths, centre position), not the box itself, so none of the layout modes move it. Zones are stored as fractions of the frame so a change of capture size does not move them either.

## Bench tools

- `localStorage['ywc.radioDisplayHotspots.debug'] = '1'` or `window.radioDisplayHotspots.debug(true)` draws every zone and the detected marker over the video.
- `window.radioDisplayHotspots.snapshot()` opens the current frame at native size for measuring.
- `localStorage['ywc.radioDisplayHotspots.layout']` takes a JSON layout override, no rebuild needed.

## Known gaps (see `docs/design/radio-display-hotspots.md`)

- One layout only: FTdx101 MONO W/F. EXPAND / MULTI / dual MAIN+SUB / 3DSS reshape the box and none of those is readable over CAT. Detecting the box from the frame is probably the right answer; a layout selector is the fallback.
- Marker colour assumed red - only seen in colour scheme 1.
- FTdx10 / FT-710 unmeasured; the module returns quietly with no layout.
- Roofing filter ring includes the optional filters; server warns rather than errors if one is not fitted.

## Other changes

- `/api/cat/status` also returns `att`, `ipo`, `agc` and `activeVfo` (additive).
- `RadioScopeControl` gains `cyclePlacement()`, `cycleSpan()`, `toggle3dss()`, `toggleHold()`.
- The pop-out page now always opens its hub connection (it was conditional on the scope dialog) and forwards updates to the overlay.

## Verification

Build and the 172 app tests pass. **Nothing here has been tried against the radio yet** - the zone geometry, the marker scan and the cycling all need a bench check.


mm5agm:
Bench update (FTdx101MP, MONO W/F layout, 2026-09-12):

- Zone table lines up with the real frame (debug overlay checked).
- Click-to-tune works in CENTER placement; SPAN, 3DSS, HOLD, ANT, ATT, IPO, AGC all work.
- Three findings fixed in 6f40e86: CURSOR needed several clicks (scope state not loaded at first click), R.FIL stuck on the unfitted 1.2 kHz optional filter (now skips on round the ring), and MONO/EXPAND looked dead (they have no CAT command; the label now flashes to say so on click).

Still open: only the one layout is measured, marker colour assumed red, FTdx10 unmeasured - see docs/design/radio-display-hotspots.md.

mm5agm:
Re-tested on the FTdx101MP after 6f40e86: CURSOR, R.FIL, and the MONO/EXPAND message all work now. Every hotspot in the table is bench-confirmed on the MONO W/F layout.

Remaining gaps are the ones in docs/design/radio-display-hotspots.md: other layouts (EXPAND/MULTI/dual/3DSS), marker colour in other colour schemes, and the FTdx10.

fabiojvalente:
@mm5agm is this limited only to the 101 for now ? 
I cant test it with the ftdx10

mm5agm:
@fabiojvalente Yes, the '101 only for now - but only because I haven't got an FTdx10 to measure the screen on, not because anything else is model-specific.

The module needs a table of where things sit on the screen, as fractions of the captured frame. It only has one, measured on the '101 in the MONO W/F layout, and for any other model it quietly does nothing - so on your FTdx10 the Radio Display page looks exactly as it did before. Everything underneath already works on the FTdx10 (scope control over CAT, the ANT/ATT/IPO/R.FIL/AGC endpoints, its own roofing-filter ring).

If you fancy measuring it, you can test without a rebuild. There is a localStorage override that is checked before the model match:

1. On the Radio Display page, open the browser console and paste:

```js
localStorage['ywc.radioDisplayHotspots.debug'] = '1';
localStorage['ywc.radioDisplayHotspots.layout'] = JSON.stringify({
  readouts: {
    ant:  [0.003, 0.416, 0.196, 0.476],
    att:  [0.199, 0.416, 0.392, 0.476],
    ipo:  [0.395, 0.416, 0.591, 0.476],
    rfil: [0.594, 0.416, 0.787, 0.476],
    agc:  [0.790, 0.416, 0.986, 0.476]
  },
  scope: {
    plot:   [0.003, 0.520, 0.986, 0.812],
    marker: [0.003, 0.540, 0.986, 0.720]
  },
  buttons: {
    cursor: [0.003, 0.847, 0.119, 0.902],
    span:   [0.122, 0.847, 0.242, 0.902],
    dss3:   [0.245, 0.847, 0.364, 0.902],
    mono:   [0.367, 0.847, 0.484, 0.902],
    multi:  [0.487, 0.847, 0.606, 0.902],
    expand: [0.609, 0.847, 0.729, 0.902],
    hold:   [0.732, 0.847, 0.851, 0.902],
    memch:  [0.854, 0.847, 0.971, 0.902]
  }
});
```

   then reload. That is the '101 table, so it will be wrong on your screen, but the debug overlay draws every zone (green = readouts, yellow = buttons, cyan = scope plot, magenta = the rows scanned for the VFO marker, dashed white = the marker it found) so you can see how far off it is.

2. `window.radioDisplayHotspots.snapshot()` opens the current frame at native size in a new tab for measuring. Each zone is `[left, top, right, bottom]` as a fraction of the frame width/height. Edit the numbers, reload, repeat until they line up.

3. Put the radio in MONO W/F with CENTER placement for click-to-tune - CURSOR/FIX work too as long as the red marker line is on screen; 3DSS is refused.

4. Paste the corrected table back here and it becomes `LAYOUTS.FTdx10`.

The other thing worth checking is the marker colour - the scan looks for saturated red, which is right for the '101 in colour scheme 1. If your FTdx10 draws the VFO line another colour the dashed white line will not appear and the hover readout will say "marker not found"; that just needs a different threshold, so tell me what colour it is.

`localStorage.removeItem('ywc.radioDisplayHotspots.layout')` puts things back.

fabiojvalente:
Am currently enabling it on the ftdx10 for testing, excited!

fabiojvalente:
@mm5agm Really like this, especially click-to-tune on the scope. Hover, click, VFO goes there — that’s the bit that makes the captured screen actually useful instead of just a picture.

I’ve put the FTdx10 overlay on a separate branch (feature/radio-display-hotspots-ftdx10) so it doesn’t land on top of this PR. MONO W/F table is in there: no ANT, SPEED instead of HOLD/MONO/MEM CH, EXPAND cycles L/N/S.

Two things that showed up on the 10:

Capture size vs what the radio is actually sending. Fit/Fill in the browser is fine. The messy case is when the dongle isn’t the same size as EXT MONITOR PIXEL — e.g. radio at 800×600, capture at 1360×768. If the stick stretches the picture, the boxes still roughly sit on the controls (picture looks squashed). If it letterboxes, the boxes include the black bars and drift. I measured at 800×600 native. Don’t assume a 16:9 pin with bars will line up.

Expanded (L) mode. On the 10 we know that over CAT, but the overlay still treats the button row as clickable after those buttons have disappeared. Ghost clicks. Same for 3DSS. MULTI we still can’t see over CAT.

The table is the measurement you asked for. Those two can wait.

As for the marker here's a screenshot of how it loooks in the 10, it is Red in every color scheme

image

mm5agm:
Thanks Fabio - merged your FTdx10 branch into the PR branch as-is (your commit, your measurements) and then reworked the module on top of it, because your findings showed the real problem: the same soft-key means a different thing on each radio. EXPAND has no CAT command on the '101, cycles L/N/S on the FTdx10, and is EXPAND/NORMAL on the FT-710 - so the zone table now says what each box *does*, not just where it is. Each zone is `{ rect, action, hint, hideWhen }`, and the model is only looked at once, to pick the table. Commit `f7e7dd4`; the design note in `docs/design/radio-display-hotspots.md` has the detail.

Your ghost-click case is handled by `hideWhen: ['size:0', '3dss']` on the FTdx10 button row - those zones stop responding (and draw dashed in debug) whenever the scope state says the row isn't on screen.

Your existing localStorage override still loads unchanged: rects come from the override, actions from the built-in table, so a box-only tweak never loses behaviour.

None of this has been on a radio since the rework. I'll re-check the '101MP; if you get a chance, the FTdx10 in L and in 3DSS is the case I can't test here.

Two other things. I had v2.5.0 ready to go before this came up, and I'm now wondering whether to hold it a few days and ship the hotspots in it rather than straight after - depends on how the two checks above go, so no pressure either way.

And I emailed you a couple of days ago about the credit for your work in the README - did that reach you? I'd like your OK on the wording before it goes out with the release.

mm5agm:
Fabio - the v2.5.0 README draft is here for you to read before I cut the release: https://gist.github.com/mm5agm/ede72525787deef2f15b305a0d90a72d

The release-notes intro and the Radio Display / Remote Audio / Docker entries are where your name appears - please check I've credited things the way you'd want. Nothing is committed yet, so any change is easy.

mm5agm:
Testing on the FTdx101MP is complete - the reworked overlay (f7e7dd4) checks out here. The only thing left before this goes in is your FTdx10 check of the overlay in size L and in 3DSS, to confirm the hideWhen zones behave. Once you've confirmed that I'll merge it and cut v2.5.0.

fabiojvalente:
@mm5agm 

**With Expand ON**
**Size S**
image

**Size N**
image

**Size L**
image


**Without Expand**

**Size S**
image

**Size N**
image

**Size L**
image



At least in the 10, the bottom buttons are always shown, we should not hide them ever.

What is the reason for a separate plot and marker areas ? - these probably need to be resized based on the 

I've also just figured out on the 10 that L and S modes in the scope controls are switched (L shows a smaller waterfall when if should be the large one - might need a fix)

mm5agm:
@fabiojvalente Thanks - those six pictures answer more than I asked. Fixed in 56f0cab and d699657:

**Button row.** You're right, it never leaves the screen. The `hideWhen` on the FTdx10 row is gone; nothing is hidden on the 10 any more. The mechanism stays in case another radio needs it.

**EXPAND.** I had it wrong on the 10 - I'd recorded it as cycling L/N/S over CAT. Your pictures and the FTdx10 manual (p26) say it's the same vertical expand as on the '101: it grows the scope up over the ATT/IPO/R.FIL/AGC row, and there's no CAT command for it. L/N/S is the waterfall touch (p25), which is what the YWC scope panel drives. The hotspot now says "no CAT command - press it on the radio" like the '101's does.

**Plot vs marker.** The marker strip is the rows I scan for the red VFO line; the plot is the whole clickable area. They're separate because the waterfall can hold red streaks in the hotter colour schemes, so I only look in the spectrum part. The strip on the 10 (0.433-0.533) stays inside the spectrum in all three sizes in your pictures, so size doesn't need to move it. What does move things is EXPAND: with it on, the scope covers the readout row, the readout zones sit over scope, and the plot's top edge is wrong. `SS` reports W/F vs 3DSS, placement and size - never EXPAND - so the table can't follow it. Same limitation on the '101, now written down in Known gaps. For now the overlay assumes EXPAND off.

**L and S.** Not switched - it is what Yaesu calls L. The FTdx10 CAT manual has `SS06` value 3 = W/F CENTER (L), 4 = (N), 5 = (S), and that is exactly what the scope panel sends. I ran the same three settings on my '101MP today: L gives the biggest *spectrum* and the smallest waterfall, N in between, S the smallest spectrum and the biggest waterfall - the same as your FTdx10 pictures. So the two radios agree with each other and with their CAT manuals; the size that Yaesu labels L is the spectrum, not the waterfall. I will leave the labels as the manual has them rather than invent a swap, but I will say so in the hover text so the next person is not surprised. On the '101 the marker strip stays in the spectrum in all three sizes too.

If you get a chance: with the fix, is anything still hidden or still clickable that shouldn't be? And 3DSS - none of the six pictures is 3DSS, and my note from your first test said the row disappears there too. Does it?

fabiojvalente:
> L gives the biggest spectrum and the smallest waterfall, N in between, S the smallest spectrum and the biggest waterfall - the same as your FTdx10 picture

This is what I got wrong, assumed it was the waterfall, thanks for clearing that one :)

Tested and all good from here 👍

fabiojvalente (approved):

mm5agm:
@fabiojvalente 
Thanks Fabio. I'm going to cut v2.5.0. now. I think that between us we have done a remarkable job. I can't think of anything else to add to YWC so I'm going to look at improving the user interface. I've already started that on IWC as it has only about 25 users but if you are interested in designing and working on the GUI I can transfer that work to YWC.  
Colin

PR #139, opened 2026-09-13
Fix FTdx10 scope UI not following front-panel SS changes

fabiojvalente:
## Summary

On FTdx10, CAT scope **writes from the UI reach the radio**, but **front-panel changes never update the buttons**. Same Controls panel on FTdx101MP/D follows SPAN/mode immediately.

**What was happening.** Live sync is not polling. The radio announces a front-panel scope change as an unsolicited `SS` frame (`AI1`). `CatMessageDispatcher` turns that into a `ScopeSetting` SignalR push and `radio-scope.js applyRemote` repaints the one sub-command that changed. That path was measured on a 101.

The FTDX10 CAT manual (ENG 2308-F) marks `SS` the same way: Set/Read/Ans/**AI all O**. P1 is `0: Fixed` (no MAIN/SUB). So the 10 is specified to use the same auto-info mechanism — it is not a radio that “doesn’t announce scope.”

Two software holes were dropping those frames before the UI saw them:

1. **Parser** — the dispatcher required a 10-character frame including `;`. A stripped 9-character body (`SS0510000`) vanished. That looks exactly like “the radio never announces.”
2. **Band tag** — `applyRemote` ignored any update whose band was not the one currently shown. On a single-receiver radio a non-`main` P1 never updated the buttons.

**The fix.** Same live-sync path as the 101 — no poll. `ScopeCommands.TryParseAnnouncement` accepts on-wire, stripped, and short `SS` bodies. On models without per-receiver scopes, the dispatcher always tags the frame MAIN; the UI applies it even without a MAIN/SUB row.

## Test plan

- [ ] **FTdx10 — UI → radio (unchanged):** change SPAN / mode / SPEED in Controls; radio TFT follows; buttons stay on the read-back.
- [ ] **FTdx10 — radio → UI (the bug):** leave Controls (or Radio Scope) open, change SPAN on the radio; highlighted button follows immediately, not after a delay. Repeat with display mode and SPEED.
- [ ] **FTdx10 — CAT log:** while turning SPAN, unsolicited `SS05…` should appear and the UI should move with it.
- [ ] **FTdx101MP/D — no regression:** front-panel SPAN still follows immediately; MAIN/SUB still independent; writes still read-back as before.
- [ ] Unit tests: `dotnet test Tests/YaesuWebControl.Tests/YaesuWebControl.Tests.csproj` (`TryParseAnnouncement`).


mm5agm:
Bench-tested on the FTdx101MP, 155/155 tests pass.

Front panel -> UI: SPAN, mode and SPEED (S.MENU) all follow immediately on MAIN and on SUB, and changing one receiver leaves the other's row alone. UI -> radio: SPAN and mode on both receivers, TFT follows, buttons stay on the read-back. HOLD unchanged from #120. Ten quick SPAN steps each way on each receiver: the event log shows every ScopeSetting in order with the right band tag and nothing dropped.

One observation for the record, not a fault: when SUB's span is changed on the front panel the '101 announces MAIN's (unchanged) span a few ms before the SUB frame, every time. Changing MAIN announces MAIN only.

No regression on the '101MP from my side. The FTdx10 items are yours to tick.

Issue #37, opened 2026-06-16
[BUG] Setting RF power to a value greater than 100 W for FTdx10


SP3L-Jacek:
It is possible to move the power slider past 100 W and it stays there presenting power greater than 100 W for the radio incapable of delivering it.

## Steps to reproduce

1. Set the RF Power to 5 W (arbitrary value).
2. Move the RF power slider to a value like 150 W. YWC allows you to do that. 150 W is now displayed as the RF Power.
3. Check the RF power in your FTdx10. It is still 5 W!

## Expected behaviour

When you slide the slider to 150 W for FTdx10, the slider should jump back to exactly 100 W, and such a power setting should be sent to the radio.

## Actual behaviour

Slider stays at 150 W and no power change is observed in the radio.

## Diagnostics

```
YWC Version    2.3.7 (released 2026-06-15)
Radio model    FTdx10
Band plan      Region1
Serial port    COM7 at 38400 baud
Connection     connected
SDR device     (none configured)
DX cluster     (off)
Cluster login  (blank)
Browser        Microsoft Edge 149
.NET runtime   10.0.8
Operating sys  Microsoft Windows 10.0.26200
CPU            Intel(R) Celeron(R) N5095 @ 2.00GHz, 4 logical cores
Memory         15.8 GB
```

## Anything else?

An alternative expected behavior is: when a 100 W radio is configured in the YWC settings, the power slider range is reduced from 200 W to 100 W.

mm5agm:
Jacek — thanks, confirmed and fixed on develop (commit 0e6003e). Cause: the slider's max-power cap only named the two FTdx101 variants explicitly, so every other 100 W model — FTdx10, FT-710, FTDX3000, FT-991A — fell through to a 200 W default.

The fix introduces a single `modelMaxPower()` mapping used by both the global and IIFE-scoped `updatePowerSliderMax()` paths:

- FTdx101MP → 200 W
- FTdx101D / FTdx10 / FT-710 / FTDX3000 / FT-991A → 100 W

Also clamps the slider's current value if it ever exceeds the new max (so changing radio model in Settings during a session pulls the slider back down rather than leaving it at the old position).

Will ship in the next release. If you'd verify on your FTdx10 once it's out, that would close this one off.

73, Colin
MM5AGM

SP3L-Jacek:
Fix in version 2.3.8 approved.

PR #140, opened 2026-09-13
Themes: an optional dark Instrument theme alongside Classic

mm5agm:
@fabiojvalente - this is a draft, opened so you can see it early rather than
because it is ready to merge. Nothing here is urgent and nothing is decided;
if you think the whole idea is wrong, say so and it goes away.

## What it is

An optional dark theme, picked in Settings > Appearance. Classic is the
default and is exactly what YWC looks like today. Instrument is dark, higher
contrast, and gives the smallest buttons a 44px hit target.

No control moves. Nothing is added to or removed from any page. A theme
changes appearance only - which is worth saying plainly, because the thing
this is NOT is the layout rework discussed in
`docs/design/ui-modernisation.md`. That one is being tried in Icom Web
Control first, where there are no collaborators and 24 downloads to break.

## The claim to check first

Classic is not "believed unchanged" - it is unreachable. Every rule in both
Instrument stylesheets is prefixed `[data-theme="instrument"]`, so with no
such attribute on `<html>` not one selector matches. If you want to falsify
that, `grep -c '^\[data-theme' ` on the two sheets against their total rule
count is the quickest route.

There is one deliberate exception, in `wwwroot/css/site.css`: the
`.frequency-display:focus` rule now reads its three values from tokens whose
`:root` defaults are the exact literals it used to contain (`#0d6efd`, `2px`,
`2px`). Identical output, and it means a theme can raise that ring's contrast
without a second rule sitting beside the first.

## Most of it is in core/

Both apps are Bootstrap 5.3.0 and build their screens out of the same
components, so the sharable part went to `core/` rather than here:

| file | what it is |
|---|---|
| `core/css/theme-tokens.css` | every value a theme may change. `:root` holds the Classic values |
| `core/css/theme-instrument.css` | the Bootstrap layer - cards, buttons, forms, tables, modals, focus |
| `core/js/theme/theme-boot.js` | applies the theme before first paint |
| `core/js/theme/theme.js` | the module API, plus the seam canvas code will need |

`core/` had no CSS at all before this, and `CopySharedCoreJs` copies
`core\js\**\*.js` only - so this adds a `CopySharedCoreCss` target that
mirrors it line for line. Including the bit that is easy to miss: it tops up
`@(Content)` itself, because the default `wwwroot/**` glob is evaluated
before the copy runs, and without it a clean first build 404s the sheets.

Only `wwwroot/css/theme-instrument-local.css` is YWC's own - the band
radiogroup, the VFO column, the calibration pages, the keypad and memory
dialogs. None of that would mean anything in IWC.

## Two things that are not obvious from the diff

**It rides on Bootstrap's own dark mode.** Bootstrap 5.3 has
`data-bs-theme="dark"`, and some of what that switches cannot be reached from
a CSS variable at all - the chevron on a `<select>` and the tick on a
`.form-check` are data-URI background images chosen by selector. So a dark
theme sets both attributes. Without it, selects and checkboxes end up with
dark glyphs on dark controls and it looks like a bug in the palette.

**`theme-boot.js` is a classic script, not a module, and has to stay one.**
`<script type="module">` is deferred by spec and runs after first paint, so as
a module it would show Classic and then swap - a white flash on every single
page load for whoever chose the dark theme. That is also why it is as small
as it is.

## Where the real risk is

The 396 `aria-*`, 42 `role=` and 83 `data-a11y-key` attributes in Index are
markup. A stylesheet cannot reach any of them, so a theme inherits all of
that for free. What a theme CAN silently destroy is the half made of CSS:
contrast, the focus ring, hit-target size. Four tokens carry it, each with
its WCAG clause written next to it in `theme-tokens.css`.

Two colours from the original mockups failed once the ratios were computed
rather than judged by eye, and ship corrected:

- caption grey `#6d8296` - 4.27:1 at the 10-11px it is actually used at (1.4.3 wants 4.5)
- alarm red `#e2453f` - 4.17:1, and that is the colour that says *you are transmitting out of band*

A third was found only by computing it: no plausible dark border reaches the
3:1 boundary for a UI component (1.4.11), so `--rwc-border-control` is
lighter than anyone would pick by eye and carries a comment asking the next
person not to darken it back.

I would rather you checked those four numbers than the CSS.

## Known not covered - the honest rough edge

The meters, the SDR spectrum, the waterfall and the filter scope are painted
by JavaScript into a `<canvas>`. CSS reaches none of it, so under Instrument
they keep their present light faces on a dark page - in the middle of the
main screen.

Fixing it means the drawing code reading its palette from the tokens.
`theme.js` already provides the seam (`themeToken` / `themePalette` /
`onThemeChange`) and nothing calls it yet. Left out deliberately: it touches
the gauge and spectrum pipelines, which are the parts of the frontend with
the strictest rules in `.claude/rules.md` and the least test cover, and it
has no business riding along inside a stylesheet change.

## Testing so far

Builds clean on `net10.0` (0 warnings, 0 errors), and `CopySharedCoreCss`
lands the sheets with a correct generated `.gitignore` that leaves the
tracked `site.css` and friends alone. Everything else is a browser check and
has not been done yet - I have not looked at a single page in Instrument on
real hardware.

Useful things to disagree with, if you have a view: whether a second theme is
worth carrying at all; whether 44px targets are right or just large; whether
the canvas being unthemed makes the dark theme not worth shipping until that
is done too.

Not urgent, and not blocking v2.5.0.

Generated with [Claude Code](https://claude.com/claude-code)


fabiojvalente:
Thanks for opening this early — really useful to see the direction while it’s still a draft.

I like a lot of the groundwork here: keeping Classic as the default (and unreachable unless chosen), putting the shared tokens in `core/`, the careful contrast work, and the clear split from the bigger layout rework. That foundation is worth keeping.

Where I’d gently push back is on what a “theme” is allowed to mean. Instrument currently bundles a few different things at once — dark colours, larger hit targets, and a bit of panel chrome. For me those work better as separate choices:

- **Colour theme:** light / dark / high contrast (palette and contrast only)
- **UI scale:** something the operator can set (font/control size), rather than baked into a named look
- **Layouts** (later, from the flex work): use-case arrangements — e.g. full remote, scope-assisted, tablet with a Main/Sub switch — without changing colours

On where the controls live: I’d rather keep appearance on the main operating page — a small button that opens a dialog — instead of another Settings section. Theme and scale are things you adjust while looking at the radio UI; burying them in Settings makes that feedback loop slower, and Settings is already long enough.

So I’d keep the token plumbing and the dark palette idea, but avoid shipping “Instrument” as one preset that also decides density.

mm5agm:
Thanks Fabio - that is the discussion I opened it for, and I agree with all three points.

Splitting Instrument is right. What I will do on this branch:

- **Colour theme** - Classic (today's light look), Dark, and a High-contrast; palette and contrast only, nothing else changes.
- **UI scale** - a separate control for font and hit-target size, independent of the colour choice.
- The name "Instrument" goes; the dark palette and the token plumbing in `core/` stay.

On where the controls live, you and I reached the same conclusion today from different directions. Earlier I set a rule for the Workspace work in IWC: the current GUI is called Classic and is never replaced, every other GUI has its own name, and where there is more than one the choice is made on the main page and remembered by that browser - not in Settings. IWC PR #37 now does that with a View selector at the top of the page. So for YWC: a small button on the main page opening an appearance dialog with the colour theme and scale in it, remembered per browser, with Settings holding only the default. Layouts, when they come across from the Workspace work, will be a third choice in the same place.

One question before I start: do you want to contribute to this one? If you would like to take any of it - the high-contrast palette, the scale control, the dialog - say which and I will leave that part alone and work around it. If not, no problem at all; I will carry on with it myself and ping you when it is worth another look.

mm5agm:
Rolling up the GUI suggestions from Kees ON9KVE's recent issues here so they are tracked in one place rather than five, since this is where the theme/layout work is being discussed anyway:

- #145 - sliders should respond to the mouse wheel (Chrome ignores the wheel on a range input; Firefox does not).
- #146 - VFO A header light grey and VFO B blue, matching the FTdx101's own screen.
- #147 - a lit/unlit key button in place of every two-choice dropdown (my Yaesu-style key-button branch already does this).
- #152 - a way to hide the meter block on the main page so the spectrum fits without scrolling.
- from #148 - three frequency-keypad refinements: a "zero the last three digits" button, a snap-to-preset button, and letting the keypad accept `3.630` for 3630 kHz without typing the leading zero.

I have closed those issues with a pointer back here. Kees, thank you - this is a useful list.

PR #141, opened 2026-09-14
WIP / feedback only: Yaesu-style VFO control keys

fabiojvalente:
## Status — feedback / evaluation only

This PR is **not** queued for merge. It is opened so reviewers can try the
UI in a browser, comment on interaction patterns, and decide whether the
direction is worth pursuing later.

There is **no short-term plan to land this on `develop`**. Treat it as a
design spike with a working branch, not a release candidate.

## Summary

Replaces a large set of classic Bootstrap selects / button groups on the
main VFO panels with Yaesu-front-panel-style keys (`ToggleDropdownButton`,
`ToggleSliderButton`, `ToggleButton`, `CycleContextButton`).

Notable behaviour in this branch:

- **Band / Mode / Segment / Roofing** — menu-only keys (click opens menu)
- **IPO** — left-click cycles IPO → AMP1 → AMP2 (no LED); right-click still picks
- **ATT / AGC / NB / Notch / Contour / APF** — toggle and/or slider keys as appropriate
- **DNR** — cycle + context; face shows **DNR/NR** when off (not “OFF”)
- **IF Width** — cycle through discrete widths; right-click slider indexes the same list; left-click resets to default (with IF Shift)
- **Band face sync** — Band key follows frequency (fixes stale “160m” while on 40m, etc.)
- USER_MANUAL / Labels updates for the new controls

image
image
image

## Out of scope / known caveats

- Visual and interaction polish still expected after feedback
- Not fully covered by automated tests (UI / live CAT paths are manual)
- Branch may need a rebase onto current `develop` before any future merge attempt

## Test plan

- [ ] Hard-refresh Index; confirm Yaesu keys render for VFO A (and B if shown)
- [ ] Band follows QSY (e.g. 7.200 MHz → **40m**, not a stale band)
- [ ] Mode / Segment / Roofing menus select and round-trip via SignalR
- [ ] IPO left-click cycle + right-click menu
- [ ] Contour / APF / NB toggle + slider
- [ ] IF Width cycle, slider, and left-click default reset; hidden in AM/FM only
- [ ] DNR off face shows DNR/NR
- [ ] Spot-check Labels page + USER_MANUAL sections touched by this branch


mm5agm:
Bench-tested this on the real radio this morning - FTdx101MP on COM4, both
RSPs streaming, VC Tune and Remote Video connected. Short version: I like it.
The layout is a real improvement and every control I tried did what it said
it would.

One note on what I actually ran. I tested `feature/yeasu-theme-with-shortcuts`
rather than this PR's `feature/ui-yaesu-style-buttons`, because that is the
branch that has everything in it - it contains this branch plus 34 more
commits (theme, shortcuts, linear gauges). Nothing in this PR's branch is
missing from it. So the feedback below covers the combined thing, and if you
would rather I retested the narrower branch on its own, say so.

Build was clean, 0 warnings, and 201/201 tests pass.

## What works

Everything I tested behaved as expected: the Yaesu-style keys (left-click
toggle, right-click menu or slider), AGC / IPO-AMP / ATT / DNR / NB / Notch /
Contour / APF, the cycle buttons, and the keyboard shortcuts. The `?` help
dialog is a nice touch and made it easy to work through them. No crashes, no
stuck states, nothing I had to go to the radio's front panel to undo.

## The one real problem: it needs a lot of screen

My monitor is 3440x1440 at 100% Windows scaling, which is a big screen, and
the page still opens at 25 percent browser zoom to fit without scrolling. At
that size I cannot read it. Working up from there:
In the following screen shots you can tell which is full screen by the lack of scroll bars.

- **25 percent** - whole page visible, no scrolling, but far too small for me to read.
image

image

- **33 percent** - I can just about read everything, but I have to scroll to
  see VFO A and VFO B. I could live with this.
image

image


- **50 percent** - perfect for reading, but I have to scroll a lot to see the
  whole app.
image

image


**One thing that is not yours, but shows up here first:** in full-screen mode
you cannot scroll, so at 50 percent the VFO A and VFO B panels are completely
unreachable - below the fold, no scrollbar, no way to get to them.

I first took that for a layout problem, and it is not. It is our own CSS and it
predates this PR. `wwwroot/css/site.css` on develop already sets
`height: 100vh !important` and `overflow: hidden !important` on
`body.fullscreen-mode`, which pins the page to viewport height and suppresses
overflow, so nothing below the fold can ever be reached. Your branch is simply
the first layout tall enough at a readable zoom to expose it - until now the
page fitted, so nobody hit it.

I have raised that separately as #163. Nothing for you to do here.

## A thought, not a request

Looking at the screenshots, there is a wide empty gap in the operating-controls
row between the A/B and Split column and the ATU / CW / DX Spots columns. On an
ultrawide that is a lot of unused horizontal space while the page is scrolling
vertically. If some of the stacked panels could spread into it the page would
get shorter, which is the thing actually causing the scrolling.

Separately, browser zoom is doing the job for me today (Ctrl and plus,
remembered per site), but this is really the same need as the UI scale control
we discussed in #140. Worth keeping the two in mind together rather than
solving it twice.

None of this changes my view that the layout is good and I want it.

mm5agm:
Full-screen scrolling is fixed - that was #163, now on develop as b90b984.

It was not anything your branch did. body is the element that goes
full screen (site.js calls body.requestFullscreen), so it is the only
scroll container there is, and it carried `overflow: hidden !important`.
Anything below the fold was simply unreachable. Your layout is just the
first one tall enough at a readable zoom to push the VFO panels past the
bottom, so it was the first time anyone hit it.

Measured on Index at 1280x720, in real full screen with real wheel input:
before, the wheel scrolled 0px of the 754px available and the VFO A/B row
sat at y=967; after, it scrolls the full 750px and that row lands at y=217.
Index, Settings, Calibrations and Memories all scroll their full extent and
the sticky header still holds at top:0. I checked it on your branch
specifically, with b90b984 cherry-picked on top of 06159a0, not just on
develop - your CSS does not collide with it.

One thing worth knowing if you are styling this app: I also removed about
100 lines of dead full-screen CSS. Index sets ViewData["BodyRole"] =
"application" for screen readers, so its <main> is role="application", and
every `main[role="main"]` rule in that block had never once matched on the
main control panel page - including an earlier `overflow-y: auto` that was
itself trying to fix this same bug. Target `main` by element rather than by
role, or it silently stops matching the moment a page sets its own.

Horizontal overflow at narrow widths is still there and I left it alone
deliberately - it predates this, the page overflows 1280px in normal mode
too, and making it scroll gives Settings a scrollbar for a 4px Bootstrap
margin artefact. Separate question whether it is worth doing.

fabiojvalente:
@mm5agm will be closing this PR and eventually start another issue and PR to track progress on the ui

PR #150, opened 2026-09-16
Add Index keyboard shortcuts for navigation and help

fabiojvalente:
## Summary
- Adds a centralized Index-page keyboard shortcut dispatcher (`keyboard-shortcuts.js`) with bindings for frequency tune, VFO/band/mode, passband, spectrum/scope, panels, and remote audio.
- Pressing `?` (or `h`, or the top-bar **?** button) opens a help dialog generated from the same registry; while open, **↑/↓** scroll the dialog (they do not tune/IF-shift). USER_MANUAL §13 documents the bindings.
- Frequency keypad via `g`: Enter sends QSY and closes; Enter/Backspace work even if focus left the keypad.
- Remote Audio mute is **Shift+M** (Space stays free for the optional TX PTT shortcut).
- **t** (or the top-bar **SDR** / **Scope** button) toggles whether **z/w/s** (and related keys) drive the SDR spectrum or the Radio Scope CAT controls. The help dialog shows a two-column SDR vs Scope table and highlights the active target.
- Fine tune: **Ctrl** + j/i/←/→ = 1 Hz; **Ctrl+Shift** + j/i/←/→ = 50 Hz (matched via `e.code` so modifiers work on macOS).
- Radio Scope: **z/Z** span (Alt = extreme), **w/W** pane size L/N/S (Alt = extreme), **o/O** reference level ±1 dB (Alt = ±10), **s** hold, **S** reset level to 0 dB. Scope actions click the same CAT-wired buttons as the panel UI.

## Test plan
- [x] `dotnet build Yaesu_Web_Control.csproj -c Debug -f net10.0`
- [x] `dotnet test Tests/YaesuWebControl.Tests/YaesuWebControl.Tests.csproj -c Debug --filter FullyQualifiedName~KeyboardShortcutRules`
- [x] On Index: press `?` — help lists shortcuts; ↑/↓ scroll; Esc closes; no frequency/IF-shift change while open
- [x] Help Spectrum/Scope table highlights current target; press **t** in help and confirm highlight / badge update
- [x] `g` → type a frequency → Enter: radio QSYs and keypad closes; out-of-range stays open with error
- [ ] Confirm shortcuts are ignored while typing in CW Send / DX Watch / text inputs
- [x] **Shift+M** mutes/unmutes Remote Audio; Space still works as TX shortcut when configured
- [x] **Ctrl** + j/i steps 1 Hz; **Ctrl+Shift** + j/i steps 50 Hz
- [ ] With both SDR and Radio Scope: **t** switches targets; choice persists across reload
- [x] On Scope target: **z/Z**, **Alt+z/Z**, **w/W**, **Alt+w/W**, **o/O**, **Alt+o/O**, **s**, **S** — confirm span / size / level / hold on the radio (and that **Shift+</>** only changes span, not level)
- [x] On CAT-only with Radio Scope: z/w/s/o drive scope without needing SDR
- [ ] Spot-check: mode letters, `m` memories, `D` DX spots, `n`/`N` VFO, band `b`/`B`
- [ ] Bench-check CAT actions (tune/mode/band/scope span+size+level+hold) against the radio


fabiojvalente:
~@mm5agm still wip, ready to test but not final set of shortcuts, as I'm considering having a toggle to switch shortcuts between scope and sdr modes~

Ready for review and feedback

mm5agm:
Bench results from `fc259c0` on the FTdx101MP (Windows 11, Edge, UK keyboard). I ran it against the rig rather than reading the diff, so this is what the keys did, not why.

**Working as described in §13**

- `h` opens the help dialog; `g` opens the keypad on the active VFO with the current frequency loaded, Enter QSYs the radio, Esc cancels
- `n` toggles the active VFO; `N` copied A → B (log: `VFO copy AB completed`) - I'd read it as "select B" at first because both VFOs were already on nearly the same frequency, so it looked like nothing happened
- `Shift+M` mutes Remote Audio; `m` opens Memories; `D` toggles DX Spots
- With Radio Scope as the target: `w` steps the pane L → N → S, `s` toggles Hold (Slow1 ↔ Hold), and the `SS1…` writes went to the SUB scope because VFO B was active - which makes sense

**Not working / questions**

1. **`?` does nothing.** `h` opens the dialog and `Shift+M` works, so Shift itself is being seen - it's specifically the `?` key that's missed. UK layout, `?` is Shift+/ as on US.
2. **Mode letters.** `l`/`u`/`c`/`q` only changed anything when the mode dropdown was already open, and the radio did not follow. The log shows the CAT going out (`MD01;`, `MD02;`, `MD04;`, `MD03;` for Receiver A between 11:58 and 12:00), so the command path is reached. Picking a mode from the dropdown with the mouse does not move the radio on this build either, and that works on develop, so this looks like a regression rather than a shortcut problem - I would treat it as the one to look at first.
3. **`Ctrl+j` / `Ctrl+i` (1 Hz) do nothing.** `Ctrl+←` / `Ctrl+→` are grabbed by the graphics driver on this PC (display hotkey), so those two chords never reach the browser here - not your bug, but worth a note in the manual that they may be taken on Windows.
4. **`t`** - I couldn't see it do anything, though the keys were clearly driving the Radio Scope. I may just have missed the SDR/Scope button flipping; a bigger cue (or the help dialog highlighting the current target, which §13 says it does) would help.
5. **`z`** worked only some of the time on the scope; **`o`** just flashed the "W/F Center - 1k - Slow1" bar with no visible level change on the radio.

Happy to re-run anything once you've had a look.

fabiojvalente:
Thanks for the bench notes — most of these were real matching/targeting bugs on our side, not “keyboard layout mystery.”

| Report | Fix |
|---|---|
| `?` does nothing (UK Edge) | Help now matches physical `Slash`+Shift (and `key === '?'` / `h`). Unshifted `/` alone restores IF width. |
| Ctrl+j/i and Ctrl+arrows “do nothing” | Chromium Downloads / GPU OSD often steal those chords. Fine tune moved off Ctrl: **`[` / `]` = 1 Hz**, **`Shift+[` / `]` = 50 Hz**. j/i ladder unchanged (100 Hz / 1 kHz / 10 kHz / DX hop). |
| `<` / `>` fragile | Span also matches `Comma`/`Period`+Shift via `e.code`. |
| Caps Lock mutes on bare `M` | Mute is **`KeyM`+Shift only**; bare `KeyM` opens Memories. |
| `t` hard to notice | Toolbar label is now **`Keys → SDR` / `Keys → Scope`**, with a short pulse on toggle; button re-syncs when an SDR panel becomes visible. |
| `z` intermittent / `o` only flashes | Scope keys no longer click buttons inside a closed dialog. One `resolveScopeControl()` picks the live host (open dialog → open/expanded card → stateful control), and **z / w / s / o / S** go through that control’s APIs only. |

Help dialog + USER_MANUAL §13 updated to match. Mode-letter / mouse MD “radio doesn’t follow” left alone for a separate CAT/session check (path unchanged vs develop).

**On repeated `o`/`O` and `z`/`Z`**

That’s expected with the current design: each keypress is a discrete CAT write (level ±1 dB / span step). Holding or mashing them floods the bus and feels laggy because every step waits on the radio (which can also trigger a race-condition like behaviour)

A better long-term fit might be a small **command palette** (VS Code–style): type the intended level or span once and send a single command, instead of stepping through every intermediate value. Happy to track that as a follow-up if you want it.

This was tested and confirmed to work on windows with a uk layout :)

Give it another pass and let me know

mm5agm:
Second pass on `eab62e5`, FTdx101MP, Windows 11 / Edge, UK keyboard. Bench only, as before.

**All working**

- `?` (Shift+/) and `h` open the help; `/` on its own restores the IF width
- `[` / `]` 1 Hz, `Shift+[` / `Shift+]` 50 Hz; `i` / `j` 100 Hz, with Shift 1 kHz, with Alt 10 kHz - checked against the dial from 21.043300
- `t` flips **Keys → SDR / Keys → Scope** and the flash makes it obvious now
- Keys → Scope: `o` / `O` level, `w` / `W` (L → N → S and back), `s` Hold
- Keys → SDR: `z` / `Z`, `w` / `W`, `s`, `S` all fine
- `Shift+M` mute; `m` Memories - and still Memories with Caps Lock on
- `b` / `B`, `n`, `N` (A → B copy), `g` → Enter QSY / Esc, `p` / `P`, `↑` / `↓` IF shift, `x`, `y`, `D`, `f`
- Mode: the dropdown moved the radio this time and the letter keys followed it, so the earlier "radio doesn't follow" was something on my side this morning, not the PR

**Two things, both minor**

1. Keys → Scope `z` / `Z` drop presses if you go fast. That matches what you said about each press being an `SS` write that waits on the radio, so I'm treating it as by design. The "type the span/level once" palette idea sounds like the right answer - yes, track it as a follow-up.
2. After the server restarted (auto-shutdown while I was away) and the tab reconnected, `z` / `Z` on Keys → Scope did nothing until I did a Ctrl+F5 - after that they worked again. Expanding the card or clicking in the panel was not needed. Reads were reaching the server but no `SS` writes. Not a blocker; mentioning it in case it's the reconnect path rather than my browser.

Typing in a text box is fine: with the cursor in the DX Watch callsign field the letters just type. If the box has not been clicked into, the letters fire shortcuts - which is right, focus is still on the page.

Happy for this to go in from my side.

